PR #26, opened 2014-10-29
Fix #25

### LINKED ISSUE #25: Unable to set Integer options to infinity
Because Inf is floating point, this code in grb_params.jl fails when trying to set something in GRB_INT_PARAMS to infinity.

```
function set_int_param!(env::Env, name::ASCIIString, v::Integer)
ret = @grb_ccall(setintparam, Cint, (Ptr{Void}, Ptr{Cchar}, Cint),
env, name, v)
if ret != 0
throw(GurobiError(env, ret))
end
end
```

IainNZ:
Is this a good fix for #25 @mlubin @joehuchette ?

### DIFF
--- a/src/grb_params.jl
+++ b/src/grb_params.jl
@@ -173,6 +173,16 @@ function set_int_param!(env::Env, name::ASCIIString, v::Integer)
     end
 end
 
+function set_int_param!(env::Env, name::ASCIIString, v::Real)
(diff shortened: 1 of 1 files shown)


IainNZ:
How do we feel about this?

mlubin:
I guess we didn't feel too good about this.

PR #34, opened 2015-03-24
Implement fixedmodel interface

odow:
Interface for the fixedmodel function in Gurobi.
http://www.gurobi.com/documentation/6.0/refman/grbfixedmodel.html

### DIFF
--- a/src/Gurobi.jl
+++ b/src/Gurobi.jl
@@ -25,7 +25,7 @@ end
 
     # grb_model
     set_sense!, update_model!, reset_model!, 
-    read_model, write_model, tune_model, presolve_model,
+    read_model, write_model, tune_model, presolve_model, fixed_model,
 
     # grb_attrs
     model_name, model_sense, model_type,
--- a/src/grb_model.jl
+++ b/src/grb_model.jl
@@ -148,3 +148,13 @@ function presolve_model(model::Model)
     end
     return Model(model.env, ret)
 end
+
+function fixed_model(model::Model)
+    @assert model.ptr_model != C_NULL
+    fixed::Ptr{Void} = C_NULL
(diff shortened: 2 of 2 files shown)


mlubin:
Looks good to me. Is the user supposed to free this model with GRBfreemodel? The documentation doesn't make this clear.

src/grb_solve.jl
```diff
@@ -18,6 +18,16 @@ function computeIIS(model::Model)
     nothing
 end
 
+function fixedModel(model::Model)
```
mlubin: This should probably go in ``grb_model.jl`` instead of ``grb_solve.jl``.

IainNZ:
Thanks for PR @odow. Is that software tool you made being used for allocating students in that 4th year sustainability class?

odow:
moved to grb_model. @IainNZ it is. You should have seen their old spreadsheet for doing it...

src/grb_solve.jl
```diff
@@ -18,7 +18,6 @@ function computeIIS(model::Model)
     nothing
 end
 
-
```
mlubin: Could you tidy this up and squash the commits?

odow:
@mlubin the user would have to free the fixed model once they are finished using it. The GRBcopy function behaves similarly in returning a new pointer.

mlubin:
@odow, ``Model`` objects are automatically freed when the julia object is garbage collected: https://github.com/JuliaOpt/Gurobi.jl/blob/eb1320cd7988b4a0aba2785001719e41c69a1bcf/src/grb_model.jl#L19

One has to be careful to not use the ``finalize_env`` option in the case where an environment is used for more than one model.

PR #38, opened 2015-05-27
fix windows v0.4- install problem

odow:
Same as https://github.com/JuliaLang/julia/commit/123e18a07c77b7576e8d2645d7c8797056f092a0


mlubin:
Thanks!

PR #39, opened 2015-07-21
[RFC] More consistent handling for constraint bounds

joehuchette:
This stores the constraint bounds in the ``GurobiMathProgModel`` and updates them whenever ``update model!`` is called. This allows us to switch equality constraints to inequality constraints (@semartin). The previous behavior didn't follow the MPB spec, since calling ``setconstrLB!(...)`` would change the right-hand side of an equality constraint, rather than only changing the lower bound.

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -7,6 +7,9 @@ type GurobiMathProgModel <: AbstractMathProgModel
     inner::Model
     last_op_type::Symbol  # To support arbitrary order of addVar/addCon
                           # Two possibilities :Var :Con
+    changed_constr_bounds::Bool # have we updated the bounds below?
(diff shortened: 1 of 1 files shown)


joehuchette:
cc @IainNZ @mlubin, this passes JuMP tests locally.

src/GurobiSolverInterface.jl
```diff
@@ -334,6 +310,8 @@ const rev_var_type_map = Compat.@compat Dict(
 )
 
 function setvartype!(m::GurobiMathProgModel, vartype::Vector{Symbol})
+    # do this to make sure we deal with new columns
+    updatemodel!(m)
```
joehuchette: Irrespective of this PR, I think this is a bugfix that our current JuMP-level behavior is masking.

joehuchette:
COOL

Issue #40, opened 2015-07-23
Bug in GurobiSolverInterface.jl/getconstrsolution


adelarue:
I've been using your wonderful Gurobi wrapper with JuMP and MathProgBase for a few months now and it's been an absolute joy, and today I think I found a small bug in getconstrsolution. When a constraint is not tight, getconstrsolution returns a value that is not within constraint bounds.
For instance, the first half of the following code fails, while the second does not:
```julia
# getconstrsolutionTest.jl

using JuMP, MathProgBase, Gurobi

# Slack GEQ constraint
m = Model(solver=GurobiSolver())
@defVar(m, x)
@addConstraint(m, x >= 0)
@addConstraint(m, x >= 3)
@setObjective(m, Min, x)

buildInternalModel(m)
im = getInternalModel(m)

MathProgBase.optimize!(im)

for i = 1:length(MathProgBase.getconstrLB(im))
	println("Checking constraint $i")
	@assert MathProgBase.getconstrLB(im)[i] <= MathProgBase.getconstrsolution(im)[i] <= MathProgBase.getconstrUB(im)[i]
end

# Slack LEQ constraint
m = Model(solver=GurobiSolver())
@defVar(m, x)
@addConstraint(m, x <= 3)
@addConstraint(m, x <= 0)
@setObjective(m, Max, x)

buildInternalModel(m)
im = getInternalModel(m)

MathProgBase.optimize!(im)

for i = 1:length(MathProgBase.getconstrLB(im))
	println("Checking constraint $i")
	@assert MathProgBase.getconstrLB(im)[i] <= MathProgBase.getconstrsolution(im)[i] <= MathProgBase.getconstrUB(im)[i]
end
```
I think this is because the variables that represent slackness of a constraint might have an overall sign, which leads to a sign error on line 264 of GurobiSolverInterface.jl. Is this correct ?

Thanks for your help !

mlubin:
This could be an issue. If you can find a bug in the logic [here](https://github.com/JuliaOpt/Gurobi.jl/blob/master/src/GurobiSolverInterface.jl#L250), PRs are accepted :)
It would be good to add this as a test to [MathProgBase](https://github.com/JuliaOpt/MathProgBase.jl/blob/0132f42277c90a866238bf18ebf8641a9b5d7466/test/linproginterface.jl) as well.

adelarue:
[...]

PR #41, opened 2015-07-24
Small update to fix issue #40

adelarue:
This answer on Gurobi Google Groups (https://groups.google.com/forum/#!topic/gurobi/8JiUTt7CcMg) suggests that Gurobi 'Slack' variables are signed based on the 'Sense' attribute of the constraint. This may have led to a small sign error in getconstrsolution which I attempted to fix. I hope this works :)

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -256,12 +256,9 @@ function getconstrsolution(m::GurobiMathProgModel)
         if sense[i] == '='
             # Must be equal to RHS if feasible
             ret[i] = rhs[i]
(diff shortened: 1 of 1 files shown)


joehuchette:
Thanks!

mlubin:
@adelarue, thanks!
Would you mind adding a small test case for this to the [MathProgBase tests](https://github.com/JuliaOpt/MathProgBase.jl/blob/0132f42277c90a866238bf18ebf8641a9b5d7466/test/linproginterface.jl)? It could just be your example from #40 (but using the ``loadproblem!`` interface instead of JuMP).

PR #45, opened 2015-10-28
add support for Gurobi 6.5

yeesian:


deps/build.jl
```diff
@@ -11,7 +11,7 @@ function write_depsfile(path)
     close(f)
 end
 
-aliases = ["gurobi60","gurobi56","gurobi55"]
+aliases = ["gurobi60","gurobi56","gurobi55","gurobi65"]
```
mlubin: Better to put it at the beginning of the list so it has priority

yeesian: right.

PR #49, opened 2016-01-28
Update grb_model.jl

gliuzzi:
There is a problem in the tune_model wrapper. A call to the gurobi function "tunemodel" does not affect the model but only computes an array of parameters settings. A subsequent call to "gettuneresult" is necessary if we want to apply the best parameters found to the given model.

Another way to accomplish this would be to encapsulate the call to "gettuneresult" within another julia wrapper.

Best wishes

### DIFF
--- a/src/Gurobi.jl
+++ b/src/Gurobi.jl
@@ -25,7 +25,7 @@ module Gurobi
     getparam, setparam!, setparams!,
 
     # grb_model
-    set_sense!, update_model!, reset_model!, 
+    set_sense!, update_model!, reset_model!, get_tune_result!,
     read_model, write_model, tune_model, presolve_model, fixed_model,
 
     # grb_attrs
--- a/src/grb_model.jl
+++ b/src/grb_model.jl
@@ -131,6 +131,14 @@ function tune_model(model::Model)
     if ret != 0
         throw(GurobiError(model.env, ret))
     end
+    #####################################################
+    # Apply best parameters setting found with tunemodel 
+    # to the given model
+    ######################################################
+	ret = @grb_ccall(gettuneresult, Cint, (Ptr{Void}, Cint), model.ptr_model, 0)
+    if ret != 0
+        throw(GurobiError(model.env, ret))
+    end
     nothing
 end
 
--- a/src/grb_model.jl
+++ b/src/grb_model.jl
@@ -131,11 +131,11 @@ function tune_model(model::Model)
     if ret != 0
         throw(GurobiError(model.env, ret))
     end
-    #####################################################
-    # Apply best parameters setting found with tunemodel 
-    # to the given model
-    ######################################################
-	ret = @grb_ccall(gettuneresult, Cint, (Ptr{Void}, Cint), model.ptr_model, 0)
+    nothing
+end
+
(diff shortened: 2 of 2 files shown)


mlubin:
@gliuzzi, the julia methods in ``grb_model.jl`` are meant to be 1-1 wrappers with the C API when possible. Could you implement a new method for ``gettuneresult``?

gliuzzi:
Of course. Could the following code be suitable?

function get_tune_result(model::Model)
	ret = @grb_ccall(gettuneresult, Cint, (Ptr{Void}, Cint), model.ptr_model, 0)
    if ret != 0
        throw(GurobiError(model.env, ret))
    end
    nothing
end

mlubin:
Yes, looks good

mlubin:
May want to take an integer argument and pass it to ``gettuneresult`` instead of zero.

gliuzzi:
Yes, agree with that. An integer specifying which one of the parameters setting one wants to use.

function get_tune_result(model::Model,i::Int)
    ret = @grb_ccall(gettuneresult, Cint, (Ptr{Void}, Cint), model.ptr_model, i)
    if ret != 0
       throw(GurobiError(model.env, ret))
    end
    nothing
end

gliuzzi:
Even better, since the passed Model gets modified by the function, maybe we should call the function with the exclamation mark at the end, e.g.

function get_tune_result!(model::Model,i::Int)
ret = @grb_ccall(gettuneresult, Cint, (Ptr{Void}, Cint), model.ptr_model, i)
if ret != 0
throw(GurobiError(model.env, ret))
end
nothing
end

mlubin:
This looks good, could you "squash" the commits from 3 into 1 by running ``git rebase -i HEAD~3``? Then force-push the branch again. This will tidy up the commit history.

gliuzzi:
Sorry, do I have to run the command "git rebase -i HEAD~3" from terminal on my machine?
I just did it. This caused an editor to open and display:

pick 260d0b9 add support for Gurobi 6.5
pick f738563 drop support for julia 0.3
pick 3c931c2 updates for MathProgBase 0.4

# Rebase d4a5ebd..3c931c2 onto d4a5ebd (3 command(s))
#
# Commands:
# p, pick = use commit
# r, reword = use commit, but edit the commit message
# e, edit = use commit, but stop for amending
# s, squash = use commit, but meld into previous commit
# f, fixup = like "squash", but discard this commit's log message
# x, exec = run command (the rest of the line) using shell
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
#
# However, if you remove everything, the rebase will be aborted.
#
# Note that empty commits are commented out

I cannot see the commits I did on the patch-1 branch of gliuzzi/Gurobi.jl  on  github.com.

Sorry, I am not used to github.

mlubin:
You should run this from the branch that you used for this pull request. Or did you make the changes from the website?

gliuzzi:
Yes, I did the changes from the website.

mlubin:
Ok, I did it manually: 6c7fd51cd341d4efcddaa1c9035e05b799d3f9e3

Issue #52, opened 2016-03-04
Segfault on ctrl-c


JackDunnNZ:
Is it possible to give Gurobi the same magic as JuliaOpt/CPLEX.jl#45 so that it doesn't segfault on ctrl-c? I'm happy to have a go if needed, but it might be easier for someone that understands the internals better than I do.

Related, I've been working with a model given to me that uses gurobipy, and on that, ctrl-c interrupts the solve and returns the current incumbent. I'm not sure if that would fall out here as a consequence, but it would be pretty cool if it worked!

mlubin:
No clue here, @joehuchette?

joehuchette:
If Gurobi has an analogue to ``cpxsetterminate``, then yeah the approach I took with CPLEX should work here as well. [grbterminate](https://www.gurobi.com/documentation/6.5/refman/c_grbterminate.html) looks like it might do the job.

(Related: https://groups.google.com/forum/#!topic/gurobi/n0CNeAe91VU)

JackDunnNZ:
Cool thanks, I'll have a crack at it.

joehuchette:
I'm guessing it'll be even easier than what I had to do in CPLEX, since you can just call ``grbterminate`` directly in ``@cpx_ccall_intercept`` instead of carrying around extra state in the model object...

JackDunnNZ:
I copied in the same macro and changed it to use `grbterminate`, but it's never catching the exception, it just dies. Any ideas? It catches the exception if I just set up an infinite loop before the solve and ctrl-c out of that, but as soon as it enters `grboptimize` it doesn't seem to catch it.

Change here, not that it's more than a copy/paste... https://github.com/JuliaOpt/Gurobi.jl/commit/16c3f714a7bf5547ead0415c98aeacaf679b151a

JackDunnNZ:
@mlubin You were right, what I have works fine as is on Ubuntu so it must be an OS X problem

PR #53, opened 2016-03-08
Fix ctrl-c causing segfault

JackDunnNZ:
Ref #52. This seems to work fine on Linux, but is patchy on OS X (I just had it work while running an unrelated solve, but most of the time, ctrl-c still causes a segfault). Not sure if you want to merge, but here it is anyway.

@joehuchette Does the CPLEX version have the same OS X problems, or is that just happening here?

### DIFF
--- a/src/grb_common.jl
+++ b/src/grb_common.jl
@@ -53,6 +53,29 @@ macro grb_ccall(func, args...)
     end
 end
 
+macro grb_ccall_intercept(model, func, args...)
+    f = "GRB$(func)"
+    quote
+        ccall(:jl_exit_on_sigint, Void, (Cint,), convert(Cint,0))
+        ret = try
+            $(@windows? :(ccall(($f,libgurobi), stdcall, $(args...))) : :(ccall(($f,libgurobi), $(args...))) )
+        catch ex
+            println("Caught exception")
+            if !isinteractive()
+                ccall(:jl_exit_on_sigint, Void, (Cint,), convert(Cint,1))
+            end
+            if isa(ex, InterruptException)
+                terminate(model)
+            end
+            rethrow(ex)
+        end
+        if !isinteractive()
+            ccall(:jl_exit_on_sigint, Void, (Cint,), convert(Cint,1))
+        end
+        ret
+    end
+end
+
 # Gurobi library version
 function getlibversion()
     _major = Cint[0]
--- a/src/grb_solve.jl
+++ b/src/grb_solve.jl
@@ -2,7 +2,7 @@
 
 function optimize(model::Model)
     @assert model.ptr_model != C_NULL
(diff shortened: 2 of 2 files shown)


joehuchette:
I haven't seen any weird behavior on the current implementation, but I definitely noticed unpredictable behavior on OSX on some previous attempts that seemed like they should've worked fine. Debugging it was beyond my pay grade, though.

src/grb_common.jl
```diff
@@ -53,6 +53,29 @@ macro grb_ccall(func, args...)
     end
 end
 
+macro grb_ccall_intercept(model, func, args...)
+    f = "GRB$(func)"
+    quote
+        ccall(:jl_exit_on_sigint, Void, (Cint,), convert(Cint,0))
+        ret = try
+            $(@windows? :(ccall(($f,libgurobi), stdcall, $(args...))) : :(ccall(($f,libgurobi), $(args...))) )
```
mlubin: Have you tried this on windows?

JackDunnNZ: Just tried it, it still solves normally. Sending ctrl-c inside cmd or powershell does nothing, and inside cygwin it segfaults.

Gurobi and JuMP tests are both fine FWIW

mlubin:
Just to clarify, what does gurobi do on ctrl-c on linux before this patch? Does it segfault?

JackDunnNZ:
Good point. I just tried it and it throws an InterruptException, but the
solve seems to keep running in the background according to CPU usage. With
the patch, it catches the exception, calls terminate, then rethrows, but
calling terminate doesn't seem to stop the solve either. I'm not sure
what's actually going on...

mlubin:
So with this patch the solver keeps running in the background?

JackDunnNZ:
Yes, it does, so as is this changes nothing. It would work if `terminate`
was changed to the appropriate command to actually stop the solver.

mlubin:
Maybe the call to ``terminate`` needs to happen inside of a callback?

JackDunnNZ:
It's possible that this is the problem, `terminate` is listed with the other callback methods, although the others explicitly say they can only be used during specific callbacks, while `terminate` says it can be called any time during the optimization (but maybe that just means regardless of the state during the callback). 

If that's the case, how do we get around it? My first guess would be to set a flag in the model to mark terminated and then set up an info callback to call `terminate` if that flag is set. I presume we can't add a new callback during a solve, but if we could then we could just add one that calls `terminate`.Any thoughts?

mlubin:
Probably worth asking on the gurobi google group to see what they suggest (and to ask why ``terminate`` doesn't work).

rdeits:
@JackDunnNZ did you ever hear back from the gurobi group about this? Being able to interrupt the solver would be extremely useful.

JackDunnNZ:
Sorry, I never pursued this further

PR #54, opened 2016-05-26
Add support for Gurobi Cloud

czlee:
I wanted to use [Gurobi Cloud](https://cloud.gurobi.com/) for a project I was working on, so I added some code to help with this. The main differences to local installations of Gurobi are:
- It uses the C function `GRBloadclientenv` rather than `GRBloadenv`.
- It takes a host name and password (to log in to the cloud server).

This code works, but it's not as neat as it could be and I'm not sure how to structure it (in a Julia-like manner) so that it doesn't repeat itself. Suggestions would be appreciated. Notice in particular how the new `Env(host, password)` constructor in grb_env.jl is a copy-paste of the existing `Env()`, but with one line changed, and ditto for `GurobiMathProgModel()` in GurobiSolverInterface.jl. Normally I'd use OOP techniques to avoid repetition but I wasn't sure how to fit it in with existing code here.

Also if you can think of a way to test it without having an actual Gurobi Cloud server running, that'd be great too. (This is why there aren't any tests.)

As hinted I don't think it's actually in mergeable condition yet, but since I'd started looking at Gurobi Cloud I thought I may as well put it forward to see if it could become something more useful. :smiley:

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -1,7 +1,7 @@
 # GurobiMathProgInterface
 # Standardized MILP interface
 
-export GurobiSolver
+export GurobiSolver, GurobiCloudSolver
 
 type GurobiMathProgModel <: AbstractLinearQuadraticModel
     inner::Model
@@ -25,13 +25,34 @@ function GurobiMathProgModel(;options...)
    return m
 end
 
+function GurobiMathProgModel(cloudhost::ASCIIString, password::ASCIIString; options...)
+   env = Env(cloudhost, password)
+   setparam!(env, "InfUnbdInfo", 1)
+   for (name,value) in options
+       setparam!(env, string(name), value)
+   end
+   m = GurobiMathProgModel(Model(env,""; finalize_env=true), :Con, false, Float64[], Float64[], nothing, nothing, nothing, nothing)
+   return m
+end
+
 
 immutable GurobiSolver <: AbstractMathProgSolver
     options
 end
+
+immutable GurobiCloudSolver <: AbstractMathProgSolver
+    host::ASCIIString
+    password::ASCIIString
+    options
+end
+
 GurobiSolver(;kwargs...) = GurobiSolver(kwargs)
 LinearQuadraticModel(s::GurobiSolver) = GurobiMathProgModel(;s.options...)
 ConicModel(s::GurobiSolver) = LPQPtoConicBridge(LinearQuadraticModel(s))
+
+GurobiCloudSolver(host, password; kwargs...) = GurobiCloudSolver(host, password, kwargs)
+LinearQuadraticModel(s::GurobiCloudSolver) = GurobiMathProgModel(s.host, s.password; s.options...)
+ConicModel(s::GurobiCloudSolver) = LPQPtoConicBridge(LinearQuadraticModel(s))
 
 supportedcones(::GurobiSolver) = [:Free,:Zero,:NonNeg,:NonPos,:SOC,:SOCRotated]
 
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -1,7 +1,7 @@
 # GurobiMathProgInterface
 # Standardized MILP interface
 
-export GurobiSolver, GurobiCloudSolver
+export GurobiSolver
 
 type GurobiMathProgModel <: AbstractLinearQuadraticModel
     inner::Model
@@ -15,8 +15,8 @@ type GurobiMathProgModel <: AbstractLinearQuadraticModel
     heuristiccb
     infocb
 end
-function GurobiMathProgModel(;options...)
-   env = Env()
+function GurobiMathProgModel(;cloudhost="", cloudpassword="", options...)
+   env = Env(cloudhost, cloudpassword)
    setparam!(env, "InfUnbdInfo", 1)
    for (name,value) in options
        setparam!(env, string(name), value)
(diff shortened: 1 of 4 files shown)


IainNZ:
Hey @czlee, thanks for the contribution! Would definitely be good to get this in.

src/GurobiSolverInterface.jl
```diff
@@ -25,14 +25,35 @@ function GurobiMathProgModel(;options...)
...
+   m = GurobiMathProgModel(Model(env,""; finalize_env=true), :Con, false, Float64[], Float64[], nothing, nothing, nothing, nothing)
+   return m
+end
+
 
 immutable GurobiSolver <: AbstractMathProgSolver
     options
 end
+
+immutable GurobiCloudSolver <: AbstractMathProgSolver
```
mlubin: I would structure this just by adding options to the existing ``GurobiSolver`` object instead of creating a new solver type.

mlubin:
Could you add tests for this? Gurobi runs a demo server at csdemo.gurobi.com according to https://www.gurobi.com/documentation/6.5/cloud-guide/client-testing.html.

czlee:
Options is a good idea, I tried it out. It looks like `options` is specifically intended for options that get passed to the Gurobi model, so I avoided piggybacking on those and used keyword arguments instead. I guess I could've popped elements out of `options` instead—let me know if you think that would be better?

Also, the cloud test only currently works if the `LD_LIBRARY_PATH` environment variable includes `${GUROBI_HOME}/lib`. The other tests work fine even if `LD_LIBRARY_PATH` is blank (I assume this part is correct behavior). The path in my `deps/deps.jl` file is correct, but I tried running `Pkg.build("Gurobi")` to try to fix dependency links and it doesn't make a difference. So I have no idea why setting `LD_LIBRARY_PATH` would make a difference to whether the cloud server (and only the cloud server!) works. Weirder still is that the error gives is 10022, [which means](https://www.gurobi.com/documentation/6.5/refman/error_codes.html) "Problem communicating with the Gurobi Compute Server". My best current theory is that, while the C function `GRBloadclientenv` might be correctly called from the shared library (using `@grb_ccall`), `GRBloadclientenv` in turn calls another function that tries to access the `libgurobi651.so` library directly, which fails (because `LD_LIBRARY_PATH` isn't set). Would you have any ideas?

IainNZ:
Thats some strange stuff, not sure what to do about it. It might just be a matter of documenting the problem.

src/grb_env.jl
```diff
@@ -3,14 +3,22 @@
...
+                a, C_NULL, cloudhost, -1, cloudpassword, 0, -1)
+        else # local installation
+            ret = @grb_ccall(loadenv, Cint, (Ptr{Ptr{Void}}, Ptr{UInt8}),
+                a, C_NULL)
+        end
         if ret != 0
             if ret == 10009
                 error("Invalid Gurobi license")
+            elseif ret == 10022
+                error("Problem communicating with the Gurobi Compute Server")
```
IainNZ: Could document the LD_LIBRARY thing here - no harm in having verbose error messages.

test/cloud.jl
```diff
@@ -0,0 +1,13 @@
+using Base.Test
+using Gurobi
+using MathProgBase.SolverInterface
+
+env = Gurobi.Env("csdemo.gurobi.com")
+Gurobi.free_env(env)
```
IainNZ: Are we allowed to solve a very small problem on the cloud server?

czlee: I'm not sure, but the [docs for the client test server](https://www.gurobi.com/documentation/6.5/cloud-guide/client-testing.html) in fact do run a small problem, so I presume it's permitted. Should we copy one from an existing test file, or encapsulate it or something and use it in both the existing test and in this cloud test, or just write a different one here?

mlubin: Yes, @cmaes suggested we solve small problems on the cloud so that we can test Gurobi.jl on Travis.

mlubin: @czlee, you can run the [mathprog](https://github.com/JuliaOpt/Gurobi.jl/blob/master/test/mathprog.jl) tests and just plugin the cloud solver.

czlee:
Thanks for the suggestions. Done in last two commits, thoughts?

The `LD_LIBRARY_PATH` thing is not ideal (because it seems to defeat the purpose of the existing scrupulousness with library referencing) but if we don't have any good answers, it's at least workable.

mlubin:
The ``LD_LIBRARY_PATH`` issue seems like a possible misconfiguration on Gurobi's part. They may be [willing](https://twitter.com/gurobi/status/736266556947206144) to help if you contact them.

test/cloud.jl
```diff
@@ -9,5 +9,5 @@ Gurobi.free_env(env)
 
 @test_throws ErrorException Gurobi.Env("fakehostdoesnotexist", "wrongpassword")
 
-solver = GurobiSolver(cloudhost="csdemo.gurobi.com")
-LinearQuadraticModel(solver)
+include(joinpath(Pkg.dir("MathProgBase"),"test","linproginterface.jl"))
+linprogsolvertest(GurobiSolver(cloudhost="csdemo.gurobi.com"))
```
mlubin: This passes? Awesome

czlee: Yup! :smile: (subject to the `LD_LIBRARY_PATH` thing)

At least most of the time... I ran into some issues initially with server capacity, but they seem to have disappeared after I removed the `LinearQuadraticModel` lines. And the issues with server capacity seemed to be caused by the test, as opposed to other users: the server reported that I was first in the queue, I would not get to the front for more than a few minutes, but once I force-stopped the test, the server seemed to have capacity again.

But I can't reproduce this issue now, it's passed the last few times I've run it.

IainNZ: Interesting. Given we aren't running tests on TravisCI, its probably OK to have possibility of flakiness when its local.

mlubin: We should be running on Travis, but it's okay if we set it up sometime later.

czlee: Oh, fair enough, I assumed lack of existing Travis was because Travis wouldn't work well with commercial software? (Just curious.)

mlubin: Well you don't need a license to use the cloud. We'd just need a script to download and set up the gurobi binaries.

ianfiske:
I think can be closed.  The latest stable release of Gurobi.jl works fine with cloud as long as the license is located in the correct place as per Gurobi's instructions.

The tests don't pass though because the test suite creates Env's iteratively without closing/freeing the old ones.  This causes the cloud pool to act as if it can't take more requests even though it's idle.

PR #55, opened 2016-05-31
warn when using callbacks outside of MIP

mlubin:
Ref https://github.com/JuliaOpt/JuMP.jl/issues/779
I'm thinking this should be a warning at the solver level instead of at the JuMP level, since Gurobi is the one silently ignoring the callbacks.
CC @leethargo @fserra


leethargo:
That makes sense. The actual reason seems to be that the solvers don't consider the root node to be `:MIPNode` iff all variables are continuous. I believe that XPress shares this behavior.

mlubin:
Does the info callback get called if Gurobi is solving an LP? Should we only warn for lazy callbacks since they are intended to affect the solution process?

leethargo:
In a quick test with `cbgetobj`, the info callback has the same behavior as a lazy constraint and is only called in the presence of integer variables.

mlubin:
Thanks for checking. The message here seems to be accurate then.

PR #56, opened 2016-07-04
Add {get, set}_{int, dbl, char}attrelement

colinfang:


mlubin:
Thanks @colinfang. Could you confirm that ``Pkg.test("Gurobi")`` passes for you locally?

colinfang:
Yep, I confirm

Issue #47, opened 2015-12-05
Segfault in JuMP tests on Gurobi 6.5


JackDunnNZ:
With JuMP/Gurobi.jl master on OSX, Gurobi 6.5 is segfaulting in the `callback.jl` tests of JuMP.

Relevant test code:

```
using JuMP, Gurobi, FactCheck
entered = [false,false]
mod = Model(solver=GurobiSolver(Presolve=0))
@defVar(mod, 0 <= x <= 2, Int)
@defVar(mod, 0 <= y <= 2, Int)
@setObjective(mod, Max, y + 0.5x)
function corners(cb)
    x_val = getValue(x)
    y_val = getValue(y)
    TOL = 1e-6
    # Check top right
    if y_val + x_val > 3 + TOL
        @addLazyConstraint(cb, y + 0.5x + 0.5x <= 3)
    end
    entered[1] = true
    @fact_throws ErrorException @defVar(cb, z)
    @fact_throws ErrorException @addLazyConstraint(cb, x^2 <= 1)
end
addLazyCallback(mod, corners)
addLazyCallback(mod, cb -> (entered[2] = true))
@fact solve(mod) --> :Optimal
```

gives

```
Optimize a model with 0 rows, 2 columns and 0 nonzeros
Coefficient statistics:
  Matrix range    [0e+00, 0e+00]
  Objective range [5e-01, 1e+00]
  Bounds range    [2e+00, 2e+00]
  RHS range       [0e+00, 0e+00]
Variable types: 0 continuous, 2 integer (0 binary)
[1]    899 segmentation fault  julia --color=yes
```

It works using Gurobi 6.0.5. @yeesian can reproduce it as well, something funny going on with callbacks?

mlubin:
``versioninfo()``?

JackDunnNZ:
```
julia> versioninfo()
Julia Version 0.4.1
Commit cbe1bee* (2015-11-08 10:33 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin13.4.0)
  CPU: Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.3
```

mlubin:
[...]

PR #57, opened 2016-07-07
Allow passing env to GurobiSolver

JackDunnNZ:
This adds the option of passing a pre-existing `Gurobi.Env` as the first arg to `GurobiSolver`. Previously `env` would be created every time the `GurobiMathProgModel` was created, which was causing issues when using Gurobi with floating-use licences if models were being created at a very fast rate. This option allows you to use the recommended setup of a single env for multiple models to get around this ([ref](http://www.gurobi.com/documentation/6.5/refman/c_grbloadenv.html))

The only other change needed is to make sure that `finalize_env` is always false if the user supplied the `env`.

I can confirm that these changes have fixed the issues I was having, and that the tests are passing for me locally modulo #47. Do we need an explicit test for this and if so what should it test?

@mlubin

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -27,7 +27,7 @@ function GurobiMathProgModel(env=nothing;options...)
 end
 function setparams!(m::GurobiMathProgModel)
     # Helper to set the parameters on the model's copy of env rather than
-    # modifying the # global env (ref: http://www.gurobi.com/support/faqs#P)
+    # modifying the global env (ref: http://www.gurobi.com/support/faqs#P)
 
     # Set `InfUnbdInfo` to 1 by default so infeasibility rays available
     setparam!(m.inner, "InfUnbdInfo", 1)
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -22,13 +22,18 @@ function GurobiMathProgModel(env=nothing;options...)
        env = Env()
    end
    m = GurobiMathProgModel(Model(env,""; finalize_env=finalize_env), :Con, false, Float64[], Float64[], nothing, nothing, nothing, nothing, options)
-   # Set the parameters on the model's copy of env rather than modifying the
-   # global env (ref: http://www.gurobi.com/support/faqs#P)
-   setparam!(m.inner, "InfUnbdInfo", 1)
-   for (name,value) in options
-       setparam!(m.inner, string(name), value)
-   end
+   setparams!(m)
    return m
+end
+function setparams!(m::GurobiMathProgModel)
+    # Helper to set the parameters on the model's copy of env rather than
+    # modifying the # global env (ref: http://www.gurobi.com/support/faqs#P)
+
+    # Set `InfUnbdInfo` to 1 by default so infeasibility rays available
+    setparam!(m.inner, "InfUnbdInfo", 1)
+    for (name,value) in m.options
+        setparam!(m.inner, string(name), value)
+    end
 end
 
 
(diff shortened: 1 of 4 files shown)


mlubin:
So there's a weird semantics issue here with parameters, since they're shared across a single environment. The following code might do something unexpected:
```jl
e = Gurobi.Env()
s1 = GurobiSolver(e, Method=2, OutputFlag=0)
s2 = GurobiSolver(e, Method=3)
m1 = Model(solver=s1)
# will this use Method = 3??
m2 = Model(solver=s2)
# will this inherit OutputFlag=0?
```

JackDunnNZ:
True, good catch. Is there an easy way to reset the environment to its default state, without recreating it, so the parameters are as expected? (I can't see one). 

Another option could be to warn that this will happen and make the user responsible for managing the parameters across different solver objects in a sensible way? The argument being that you should know what you are doing if you need this in the first place

mlubin:
According to http://www.gurobi.com/support/faqs:
> Each model gets its own copy of the Gurobi environment; this allows each model to have its own parameter values. However, this means that you must set parameters on the environment for the model, rather than on the master environment. To see how to set parameters on the environment for a model, see the source code in the examples subdirectory, such as the params example.

So we should just set the parameters on the model's copy of the environment.

JackDunnNZ:
I think this is already being done (`setparam!` on `Model` uses the
enviroment returned by `GRBgetenv` as per their example, rather than the
global env), but I will add some tests to make sure

mlubin:
The ``GurobiMathProgModel`` constructor calls ``setparam!`` on the passed-in environment directly though.

JackDunnNZ:
Whoops, okay. Looks like we can move that down a line to after the model is created and call on `m.inner` rather than `env`?

mlubin:
Yeah

JackDunnNZ:
@mlubin I've made the changes so that the options are set on the model's copy of the env. The only complication is that the model's env is re-created each time in `loadproblem!` and loses all of the old parameters. I fixed it by carrying around the options in the `GurobiMathProgModel` as well so that they could be re-added to the new model's env. Let me know if you can think of a better way.

I also added some simple tests to check that the env is passed through alright, and lives through a finalizer if we supplied it manually. 

The JuMP tests actually cover the other changes, since the model env is repeatedly rebuilt, and would otherwise lose parameters like `QCPDual` and then fail the corresponding tests.

src/GurobiSolverInterface.jl
```diff
@@ -14,35 +14,48 @@ type GurobiMathProgModel <: AbstractLinearQuadraticModel
...
-  # throw away old model
+  # throw away old model but keep env and finalize_env
   env = m.inner.env
+  finalize_env = m.inner.finalize_env
   m.inner.finalize_env = false
   free_model(m.inner)
-  m.inner = Model(env, "", finalize_env=true)
+  m.inner = Model(env, "", finalize_env=finalize_env)
+  # Re-set options on new model's env
+  setparam!(m.inner, "InfUnbdInfo", 1)
```
mlubin: Why is this set here?

mlubin: Oh, forgot this was already being set. Might be worth an in-line comment

mlubin:
Some merge conflicts crept in but looks good. Could you document this in the readme as well?

src/GurobiSolverInterface.jl
JackDunnNZ: Another option would be just to set it on the env itself rather than the model's copy?

mlubin: I'll leave that up to you

JackDunnNZ:
Sure, whereabouts do you want it? At the end of the "Gurobi Environment" section?

mlubin:
I would just add a whole new subsection at the bottom. The readme is in bad shape anyway.

JackDunnNZ:
Okay, rebased and added a note to the readme (a bit clumsy, feel free to tweak anything). I also pulled out the parameter setting into a function and added a comment there explaining why we set `InfUnbdInfo`. I think it's better to re-set this at the model-level every time since that's what was happening before this change.

mlubin:
Thanks!

PR #58, opened 2016-07-29
deprecation fixes for Julia 0.5

mlubin:
@tkelman, does this seem reasonable?

### DIFF
--- a/src/grb_model.jl
+++ b/src/grb_model.jl
@@ -21,9 +21,10 @@ end
 
 # If environment is tied to this model, the, we can safely finalize
 # both at the same time, working around the Julia GC.
-function Model(env::Env, name::ASCIIString; finalize_env::Bool=false)
+function Model(env::Env, name::String; finalize_env::Bool=false)
 
     @assert is_valid(env)
+    @assert isascii(name)
     
     a = Array(Ptr{Void}, 1)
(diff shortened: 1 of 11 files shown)


src/grb_model.jl
```diff
@@ -104,7 +106,7 @@ end
 
 # read / write file
 
-function read_model(model::Model, filename::ASCIIString)
+function read_model(model::Model, filename::String)
```
tkelman: do utf8 filenames work even on windows?

tkelman:
I usually prefer something more descriptive than assert, but mostly

mlubin:
> I usually prefer something more descriptive than assert, but mostly

I'll leave it to Gurobi to make the interface more professional when they feel like it

src/grb_model.jl
mlubin: No idea, but non-ascii filenames should work in principle

tkelman: the win32 unicode api's will usually be in utf16 though, so it depends what gurobi uses to manipulate files

mlubin: I have no motivation to spend time figuring this out and testing on different platforms. Should I add an ``isascii`` assert with a TODO?

tkelman: I think that would be safer

Issue #60, opened 2016-09-15
Confirmed upstream bug: get/set of objective coefficients


odow:
See https://groups.google.com/forum/#!topic/gurobi/R9SuKdb8YHc for the discussion.

Files demonstrating this (in both Julia and Python) are here
https://gist.github.com/odow/5255c34daef85a752da76b3836f20be3

Briefly:

After solving some LP, we add a new constraint
https://gist.github.com/odow/5255c34daef85a752da76b3836f20be3#file-main-jl-L25-L26

During the subsequent resolve (https://gist.github.com/odow/5255c34daef85a752da76b3836f20be3#file-main-jl-L30), Gurobi reformulates the model

The LP (and MPS) file of the model at this stage leaks this reformulation (minor bug).
https://gist.github.com/odow/5255c34daef85a752da76b3836f20be3#file-3_julia-jl-L1-L16

However, this leak does not occur through the other API's. Therefore, when we get the objective coefficients, we query the original coefficients. 
https://gist.github.com/odow/5255c34daef85a752da76b3836f20be3#file-main-jl-L34

When we set the objective coefficients (which on resolve, JuMP always does using the original JuMP model coefficients https://github.com/JuliaOpt/JuMP.jl/blob/master/src/solvers.jl#L356 ), we trigger a bug, overriding the reformulated coefficients with this original coefficients.
https://gist.github.com/odow/5255c34daef85a752da76b3836f20be3#file-main-jl-L35

In the next solve, Gurobi "undoes" the formulation, but doesn't account for us resetting the correct (but incorrect for that formulation) coefficients (major bug).

This creates the unboundedness.
https://gist.github.com/odow/5255c34daef85a752da76b3836f20be3#file-julia-log-L39

This is pretty worrying for any model solved with JuMP and Gurobi as JuMP always performs this supposed no-op of getting and setting the objective coefficients. If Gurobi has ever reformulated a model in a way that is not unbounded, there would be no way of knowing (short of manually recalculating the objective) that the objective coefficients were switched, leading to an incorrect solution.

@mlubin

mlubin:
Closing since gurobi confirmed it's an upstream issue

mlubin:
@odow, I'm open to suggestions on how to avoid incorrect solutions until the next gurobi release. Do you have any understanding of the potential scope of this issue?

odow:
I just tested the Beta version for Gurobi 7.0 which fixes this.

[...]

PR #63, opened 2016-09-26
Add catch for objective bug through MPB layer

odow:
See https://github.com/JuliaOpt/Gurobi.jl/issues/60

The alternative is catching this at the C interface level. However I went this (quick) route first because I presume/hope no one is using Gurobi.jl by itself. Going the C route would involve catching the setAttr functions that can be used to modify the objective as well as the add variable functions.

This bug has existed since the very first version of Gurobi. Some sort of catch definitely needs to go in. Gurobi ends up solving a different problem to the one you specify so without it we really have no idea of the frequency of this occurrence.

@mlubin

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -140,7 +140,7 @@ setconstrLB!(m::GurobiMathProgModel, lb) = (m.changed_constr_bounds = true; m.lb
 setconstrUB!(m::GurobiMathProgModel, ub) = (m.changed_constr_bounds = true; m.ub = copy(ub))
 
 getobj(m::GurobiMathProgModel)     = get_dblattrarray( m.inner, "Obj", 1, num_vars(m.inner)   )
-setobj!(m::GurobiMathProgModel, c) = (m.obj=c; set_dblattrarray!(m.inner, "Obj", 1, num_vars(m.inner), c))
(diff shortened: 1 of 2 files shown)


mlubin:
Seems reasonable to me. It's not our job to fix the C interface.

src/GurobiSolverInterface.jl
```diff
@@ -139,20 +140,22 @@ setconstrLB!(m::GurobiMathProgModel, lb) = (m.changed_constr_bounds = true; m.lb
 setconstrUB!(m::GurobiMathProgModel, ub) = (m.changed_constr_bounds = true; m.ub = copy(ub))
 
 getobj(m::GurobiMathProgModel)     = get_dblattrarray( m.inner, "Obj", 1, num_vars(m.inner)   )
-setobj!(m::GurobiMathProgModel, c) = set_dblattrarray!(m.inner, "Obj", 1, num_vars(m.inner), c)
+setobj!(m::GurobiMathProgModel, c) = (m.obj=c; set_dblattrarray!(m.inner, "Obj", 1, num_vars(m.inner), c))
```
mlubin: Maybe ``m.obj = copy(c)`` here to avoid ownership issues?

odow:
@mlubin how's it now. 

We are currently in discussion with Gurobi about whether they are going to issue a patch to cover old versions. I'm very interested to see how often this gets triggered.

mlubin:
Go ahead and tag a new release if you like. Let's see what happens.

PR #65, opened 2016-09-28
implement solver-independent parameters

mlubin:
CC @chriscoey

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -38,7 +38,7 @@ function setparams!(m::GurobiMathProgModel)
 end
 
 
-immutable GurobiSolver <: AbstractMathProgSolver
+type GurobiSolver <: AbstractMathProgSolver
     env
     options
 end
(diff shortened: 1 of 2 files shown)


src/GurobiSolverInterface.jl
```diff
@@ -48,6 +48,37 @@ ConicModel(s::GurobiSolver) = LPQPtoConicBridge(LinearQuadraticModel(s))
...
 supportedcones(::GurobiSolver) = [:Free,:Zero,:NonNeg,:NonPos,:SOC,:SOCRotated]
 
+function setparameters!(s::GurobiSolver; mpboptions...)
+    opts = collect(Any,s.options)
+    for (optname, optval) in mpboptions
+        if optname == :TimeLimit
+            push!(opts, (:TimeLimit,optval))
+        elseif optname == :Silent
+            if optval == true
+                push!(opts, (:OutputFlag,0))
```
odow: Not sure whether this should be `OutputFlag` or `LogToConsole`.
https://www.gurobi.com/documentation/6.5/refman/logtoconsole.html

odow: I guess it doesn't actually matter. Since the log file is disabled by default. So you would have to manually set the log file parameter and by that stage you probably already know about `LogToConsole` if you want to turn screen logging off...

mlubin: Yeah, if you want that level of fine grained control then you should use solver specific parameters.

PR #71, opened 2016-12-10
fix on 0.6 and drop 0.4

mlubin:


joaquimg:
Have anyone tested [this](https://discourse.julialang.org/t/macro-that-uses-ccall-in-julia-0-6-0/2639/5) problem on windows?

It seems that we need to escape the stdcall on the macros...

mlubin:
No, I haven't. Note that stdcall is only an issue for win32, doesn't affect win64. So in the worst case this affects a small number of users.

joaquimg:
Actually I just tested and Gurobi.jl does not load in win64 (in 0.6.0-alpha).

`using Gurobi`
returns:
```
ERROR: LoadError: LoadError: syntax: ccall argument types must be a tuple; try "(T,)"
Stacktrace:
 [1] include_from_node1(::String) at .\loading.jl:539
 [2] include(::String) at .\sysimg.jl:14
 [3] include_from_node1(::String) at .\loading.jl:539
 [4] include(::String) at .\sysimg.jl:14
 [5] anonymous at .\<missing>:2
while loading C:\Users\joaquimgarcia\.julia\v0.6\Gurobi\src\grb_common.jl, in expression starting on line 58
while loading C:\Users\joaquimgarcia\.julia\v0.6\Gurobi\src\Gurobi.jl, in expression starting on line 62
ERROR: Failed to precompile Gurobi to C:\Users\joaquimgarcia\.julia\lib\v0.6\Gurobi.ji.
Stacktrace:
 [1] compilecache(::String) at .\loading.jl:673
 [2] require(::Symbol) at .\loading.jl:460
```

Which is exactly the problem I am talking about in that issue above.

My best solution was:
```
macro xprs_ccall(func, args...)
   f = "XPRS$(func)"
   args = map(esc,args)

   is_unix() && return quote
       ccall(($f,xprs), $(args...))
   end
   is_windows() && VERSION < v"0.6-" && return quote
       ccall(($f,xprs), stdcall, $(args...))
   end
   is_windows() && VERSION >= v"0.6-" && return quote
       ccall(($f,xprs), $(esc(:stdcall)), $(args...))
   end
end
```

Messy but works...

mlubin:
Oh, could you make a PR?

joaquimg:
I will! just writing an issue.

Issue #73, opened 2016-12-18
Omission in Documentation?


gdowdy3:
I'm new to using Gurobi in Julia, so I was following your README, looking at the examples.  In particular, I was looking at Example 1.4.  I could run the code and solve the optimization problem, but I noticed that any parameters I attempted to manipulate through a combination of 

     env = Gurobi.Env()
and

     setparam!(env, name, v) 
had no effect on the results.

After more time than I care to admit, I realized that the problem was the following line in Example 1.4:

    m = Model(solver=GurobiSolver())
As soon as I updated this to read

    m = Model(solver=GurobiSolver(env))
My parameter manipulations had the expected effect.

For the seasoned Gurobi users out there, it is perhaps obvious that we somehow need to inform Gurobi of the changes made to the environment.  However, this was not obvious to a noob like me.  You might consider updating the troublesome line in Example 1.4 to spare future noobs the same confusion.

mlubin:
Happy to consider pull requests. We receive no support from Gurobi for maintaining this package.

odow:
There is a brief mention about sharing `env`'s right at the bottom of the readme. Maybe it just needs to be made more explicit.

@gdowdy3 you can edit the readme in Github by navigating to https://github.com/JuliaOpt/Gurobi.jl/blob/master/README.md and clicking the pencil icon on the top right of the file.

PR #74, opened 2016-12-18
Update of README file in response to Issue #73

gdowdy3:
https://github.com/JuliaOpt/Gurobi.jl/issues/73

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ A = [50. 24.; 30. 33.]
 b = [2400., 2100.]
 lb = [5., 45.]
  
-solution = linprog(f, A, '<', b, lb, Inf, GurobiSolver())
+solution = linprog(f, A, '<', b, lb, Inf, GurobiSolver(env))
 ```
 
 ##### Example 1.4: Linear programming with JuMP (Algebraic model)
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ natural algebraic approach.
 ```julia
 using JuMP
 
-m = Model(solver=GurobiSolver())
+m = Model(solver=GurobiSolver(env))
 
 @variable(m, x >= 5)
 @variable(m, y >= 45)


odow:
You don't have to pass an `env` to the JuMP model. You might just want to add the link to the part where it discusses sharing multiple `env`s (https://github.com/JuliaOpt/Gurobi.jl#reusing-the-same-gurobi-environment-for-multiple-solves) near the JuMP example.

odow:
Or you could add something like
```julia
m = Model(solver=GurobiSolver())
# you can also use an existing Gurobi.Env as follows
# env = Gurobi.Env()
# setparam!(env, name, v) 
# m = Model(solver=GurobiSolver(env))
```

odow:
While you're at it, the example should have 
```julia
using JuMP, Gurobi
```
as the first line.

gdowdy3:
@odow, I understand that I don't have pass the `env` in order to construct the model and then solve the optimization problem.  But it seems that passing the environment is necessary if I want to modify the solver options - for example to remove preprocessing.  Do you agree?

odow:
You can also pass options as keywords. i.e.:
```julia
m=Model(solver=GurobiSolver(Presolve=0))
```
Here is where they get mapped into `setparam!` calls: https://github.com/JuliaOpt/Gurobi.jl/blob/master/src/GurobiSolverInterface.jl#L36

gdowdy3:
That may be true.  The point is: the README is inconsistent inconsistent in this section.  It suggests that every example has a preamble in which the `env` is modified, but then Example 1.4 does not make use of this modified environment.

odow:
Okay now I see! How about you remove that part and just give the full preamble in each example with the example of setting `Presolve` to zero? For examples 1.1 and 1.2 it would be via the `env` and for 1.3 and 1.4 it would be via the `GurobiSolver` keyword. Thanks for clarifying this!

gdowdy3:
I feel that the README is clear enough if we just add `env` to Examples 1.3 and 1.4 as I've done.

mlubin:
I don't think we should be encouraging use of Env by default in the cases
where passing keywords to GurobiSolver works. The main use case of Env is
to manage license issues where the number of simultaneous solves is limited.

On Dec 18, 2016 16:35, "gdowdy3" <notifications@github.com> wrote:

I feel that the README is clear enough if we just add env to Examples 1.3
and 1.4 as I've done.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/JuliaOpt/Gurobi.jl/pull/74#issuecomment-267848560>, or mute
the thread

PR #78, opened 2016-12-23
New functions for MathProgBase

joaquimg:
Prepares Gurobi to be used in MathProgBase with the functions:  

copy,  delconstrs!, delvars!, chgcoeffs

Adds tests


mlubin:
Hard to review before seeing the corresponding PR to MathProgBase

src/GurobiSolverInterface.jl
```diff
@@ -35,11 +35,11 @@ function copy(m::GurobiMathProgModel)
                                copy(m.obj), 
                                copy(m.lb), 
                                copy(m.ub), 
-                               deepcopy(lazycb),
-                               deepcopy(cutcb),
-                               deepcopy(heuristiccb), 
-                               deepcopy(infocb), 
-                               deepcopy(options))
+                               deepcopy(m.lazycb),
```
mlubin: It's a bit unclear if it's safe to copy callbacks. How would this work if the callbacks hold a reference to the old model, for example?

joaquimg:
Actually I have no idea about what to do with callbacks, since I have not
been dealing with them, maybe just check if callbacks exist and tell the
user that they were not copied?

I see the copy part mostly for LP, since MIPs tend to be slow to solve and
thus rebuilding from scratch would not be a big bottleneck, but its hard to
tell...

I am finishing the MathProgBase PR, but I getting used to some few
differences in GLPK

joaquimg:
Can we move with this one?

mlubin:
Yes, but there's a merge conflict thanks to https://github.com/JuliaOpt/Gurobi.jl/pull/89

PR #79, opened 2016-12-30
Fix loadproblem! from file

blegat:
When `loadproblem!` is called with a file, `lb`, `ub` and `obj` do not correspond to the internal value of Gurobi. This PR fixes this by setting them to nothing to show that they do not represent the internal value and set them to the internal value when needed.

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -79,7 +79,12 @@ function setparameters!(m::GurobiMathProgModel; mpboptions...)
     end
 end
 
-loadproblem!(m::GurobiMathProgModel, filename::AbstractString) = read_model(m.inner, filename)
+function loadproblem!(m::GurobiMathProgModel, filename::AbstractString)
+    read_model(m.inner, filename)
+    m.obj = getobj(m)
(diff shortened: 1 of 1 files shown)


mlubin:
Could we just initialize ``lb``, ``ub``, and ``obj`` properly inside ``loadproblem!`` instead of introducing more possible states into the gurobi model?

blegat:
Yes, I did hesitate with doing that too. I thought that it could be useful in other cases to have the possibility to set them to Nullable to make it clear that they are not in sync and that we want to avoid doing useless work by syncing them since it might not be needed. If you are convinced that always syncing is preferable then we can do that

mlubin:
My philosophy is to reduce the possible number of states that an object can be in whenever reasonable. Always synced is better than sometimes synced unless there's a good reason to allow the latter.

blegat:
I see what you mean, my approach was to ensure that the MPB layer is as thin as possible so that there wouldn't be any reason to bypass it. Now since we only copy `lb`, `ub` and `obj` (and not, e.g. `A`), it may be considered negligeable.

blegat:
@mlubin I have removed the additional states :)

Issue #82, opened 2017-02-07
Support hierarchical (or lexicographic) multi-objective linear programming?


cossio:
Hierarchical linear programming is when you have more than one objective, and you optimize them hierarchically. For example, if z1 is the optimal result from the first objective, you then optimize the second objective, but adding the contstrain that the first objective function is fixed at z1, and so on.

Gurobi has an interface for this, as I gather from the manual. It would be nice if this were ported to Gurobi.jl.

Note that simply doing the two optimizations by hand seems to have numerical issues in my experience. I have found that frequently the secondary optimization is seen as infeasible.

PR #83, opened 2017-02-07
Support hierarchical multi-objectives

cossio:
Fix https://github.com/JuliaOpt/Gurobi.jl/issues/82.

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -181,24 +181,28 @@ setconstrUB!(m::GurobiMathProgModel, ub) = (m.changed_constr_bounds = true; m.ub
 getobj(m::GurobiMathProgModel)     = get_dblattrarray( m.inner, "Obj", 1, num_vars(m.inner)   )
 setobj!(m::GurobiMathProgModel, c) = (m.obj=copy(c); set_dblattrarray!(m.inner, "Obj", 1, num_vars(m.inner), c))
 
-function setmultiobj!(m::GurobiMathProgModel, i::Int, c)
-    set_intattr!(m.inner, "ObjNumber", i)
-    set_dblattrarray!(m.inner, "Obj", 1, num_vars(m.inner), c)
+function setmultiobj!(m::Model, i::Int, c)
+    set_int_param!(m.env, "ObjNumber", i)
+    set_dblattrarray!(m, "Obj", 1, num_vars(m), c)
+    set_int_param!(m.env, "ObjNumber", 1)
 end
 
-function getmultiobj(m::GurobiMathProgModel, i::Int)
-    set_intattr!(m.inner, "ObjNumber", i)
-    get_dblattrarray(m.inner, "Obj", 1, num_vars(m.inner))
+function getmultiobj(m::Model, i::Int)
+    set_int_param!(m.env, "ObjNumber", i)
+    get_dblattrarray(m, "Obj", 1, num_vars(m))
+    set_int_param!(m.env, "ObjNumber", 1)
 end
 
-function setmultiobjpriority!(m::GurobiMathProgModel, i::Int, priority::Int)
-    set_intattr!(m.inner, "ObjNumber", i)
-    set_intattr!(m.inner, "ObjNPriority", priority)
+function setmultiobjpriority!(m::Model, i::Int, priority::Int)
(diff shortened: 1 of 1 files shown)


cossio:
Where should I add tests?

mlubin:
``getobj`` and ``setobj!`` implement a common interface defined in MathProgBase. Any change in the definition of the arguments would have to take place in the MathProgBase.jl package. Secondly, since Gurobi is the only solver I know which supports this, we'd be unlikely to modify MathProgBase to support a feature only available in one solver.

In your own code, I'd recommend using ``get_dblattrarray( m.inner, "Obj", i, num_vars(m.inner)   )`` directly, or you can add helper functions to Gurobi.jl which don't conflict with MathProgBase.

cossio:
@mlubin Why do you say this conflicts with `MathProgBase`? We have a default value of `i = 1` in the method definition, so you can still call it as before.

mlubin:
It conflicts in the sense that it defines a new method called
```jl
MathProgBase.setobj!(::GurobiMathProgModel,::Vector,::Int)
```
which is entirely undocumented in MathProgBase. This could easily lead to confusion for whoever sees it being used in a piece of code. It's fine to have a method with this functionality, but it should not be called ``MathProgBase.setobj!`` (until the documentation in MathProgBase is changed).

cossio:
I followed @blegat suggestion. Now it's a separate function with no conflicts with MathProgBase.

blegat:
I have no idea whether my suggestion actually works. Have you tested it ?

cossio:
testing now! let's see.

cossio:
I tested this one (on very simple cases) and it seems to work.

cossio:
A test could be something like:

```julia
env = Gurobi.Env()
A = [1. 1.]
b = [1.]
lb = [0.; 0.]
f = [1.; 1.]
model = gurobi_model(env, name = "My Model", sense = :maximize, A = A, b = b, lb = lb, f = f)

setmultiobj!(model, 2, [0.; 1.])
optimize(model)
@assert get_solution(model) == [0.; 1.]

setmultiobj!(model, 2, [1.; 0.])
optimize(model)
@assert get_solution(model) == [1.; 0.]
```

cossio:
It's not really working, I was only changing the original objective. I am working a on a fix. Also need a better test case.

blegat:
@cossio you can add the tests in a file (say) `multiobj.jl` and then add `"multiobj.jl"` in the `tests` array in the `runtests.jl` file.

cossio:
@mlubin @blegat I am replacing this PR with https://github.com/JuliaOpt/Gurobi.jl/pull/86, which also contains tests.

Sorry for the mess, I'm still getting the hang of github.

src/GurobiSolverInterface.jl
```diff
@@ -178,35 +178,59 @@ end
...
+function get_multiobj_weight(m::Model, i::Int)::Float64
+    i0 = get_int_param(m, "ObjNumber")
+    set_int_param!(m, "ObjNumber", i)
+    w::Float64 = get_dblattr(m, "ObjNWeight")
+    set_int_param!(m, "ObjNumber", i0)
+    return w
+end
+
+function set_multiobj!(m::Model, i::Int, c, p, w)
+    set_multiobj!(m, i, c)
```
blegat: This should be `set_multiobj_c!`

cossio: @blegat No, because `set_multiobj!` sets all three, `c, p, w` (objective coefficients, priority, and weight). A different function `set_multiobj_c!` sets only `c`. Don't you agree?

cossio: Please, move the discussion to https://github.com/JuliaOpt/Gurobi.jl/pull/86.
Or, is there a way to "merge" two pull requests into a single pull request?

blegat: I mean the first line `set_multiobj!(m, i, c)` should be changed to `set_multiobj_c!(m, i, c)`

cossio: @blegat  Done. I fixed this on https://github.com/JuliaOpt/Gurobi.jl/pull/86.

cossio: I must have missed this because in the tests I ran, I had a hanging definition of `set_multiobj!(m, i, c)` on the REPL.

PR #87, opened 2017-03-13
add more status codes

omalled:
This pull request adds a couple status codes that are documented [here](https://www.gurobi.com/documentation/7.0/refman/optimization_status_codes.html). The last status code seems to be required to use the [BestObjStop](http://www.gurobi.com/documentation/7.0/refman/bestobjstop.html#parameter:BestObjStop) option like `Gurobi.GurobiSolver(BestObjStop=...)`.

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -318,6 +318,10 @@ function status(m::GurobiMathProgModel)
     return :Suboptimal # not very useful status
   elseif s == :interrupted # ended by user?
     return :UserLimit
+  elseif s == :inprogress
+      return :InProgress
+  elseif s == :user_obj_limit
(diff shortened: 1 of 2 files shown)


joehuchette:
Looks good, thanks.

src/GurobiSolverInterface.jl
```diff
@@ -318,6 +318,10 @@ function status(m::GurobiMathProgModel)
     return :Suboptimal # not very useful status
   elseif s == :interrupted # ended by user?
     return :UserLimit
+  elseif s == :inprogress
+    return :InProgress
+  elseif s == :user_obj_limit
+    return :UserObjLimit
```
mlubin: This should be ``:UserLimit``

PR #88, opened 2017-03-14
Functions for getting string array attributes

iaravena:
adds get_strattrarray and get_strattrarray!

### DIFF
--- a/src/grb_attrs.jl
+++ b/src/grb_attrs.jl
@@ -146,7 +146,7 @@ function get_strattrarray!(r::Array{Ptr{UInt8}}, model::Model, name::String, sta
     if ret != 0
         throw(GurobiError(model.env, ret))
     end
-	map(unsafe_string, r)
+    map(unsafe_string, r)
 end
 
 # setters
--- a/src/grb_attrs.jl
+++ b/src/grb_attrs.jl
@@ -133,6 +133,22 @@ function get_charattrarray(model::Model, name::String, start::Integer, len::Inte
     get_charattrarray!(Array{Cchar}(len), model, name, start)
 end
 
+function get_strattrarray(model::Model, name::String, start::Integer, len::Integer)
+    @assert isascii(name)
+    get_strattrarray!(Array(Ptr{UInt8}, len), model, name, start)
+end
+
+function get_strattrarray!(r::Array{Ptr{UInt8}}, model::Model, name::String, start::Integer)
+    @assert isascii(name)
+    ret = @grb_ccall(getstrattrarray, Cint,
+        (Ptr{Void}, Ptr{UInt8}, Cint, Cint, Ptr{Ptr{UInt8}}),
+        model, name, start - 1, length(r), r)
+    if ret != 0
+        throw(GurobiError(model.env, ret))
+    end
+	map(unsafe_string, r)
+end
+
 # setters
 
 function set_intattr!(model::Model, name::String, v::Integer)


src/grb_attrs.jl
```diff
@@ -133,6 +133,22 @@ function get_charattrarray(model::Model, name::String, start::Integer, len::Inte
...
+
+function get_strattrarray!(r::Array{Ptr{UInt8}}, model::Model, name::String, start::Integer)
+    @assert isascii(name)
+    ret = @grb_ccall(getstrattrarray, Cint,
+        (Ptr{Void}, Ptr{UInt8}, Cint, Cint, Ptr{Ptr{UInt8}}),
+        model, name, start - 1, length(r), r)
+    if ret != 0
+        throw(GurobiError(model.env, ret))
+    end
+	map(unsafe_string, r)
```
blegat: You should replace this tab character by 4 spaces :)

iaravena: true! I didn't noticed it. thanks

src/grb_attrs.jl
```diff
@@ -133,6 +133,22 @@ function get_charattrarray(model::Model, name::String, start::Integer, len::Inte
...
+function get_strattrarray(model::Model, name::String, start::Integer, len::Integer)
+    @assert isascii(name)
+    get_strattrarray!(Array(Ptr{UInt8}, len), model, name, start)
+end
+
+function get_strattrarray!(r::Array{Ptr{UInt8}}, model::Model, name::String, start::Integer)
+    @assert isascii(name)
+    ret = @grb_ccall(getstrattrarray, Cint,
+        (Ptr{Void}, Ptr{UInt8}, Cint, Cint, Ptr{Ptr{UInt8}}),
+        model, name, start - 1, length(r), r)
```
joehuchette: This will pass garbage in the third argument. You should change it to something like ``Cint(start-1)``.

joehuchette: Also, the fourth argument should be ``Cint(length(r))``.

mlubin: I think there's automatic conversion for scalars in this case

joehuchette: From Int64 to Int32?

mlubin: To the type specified in the ccall.

mlubin:
@iaravena, could you add tests for these in ``test/test_grb_attrs.jl``?

iaravena:
It's already working in my machines (tried on Win7, Win10 and MacOS). Is there anywhere else where could I try it? (as far as I know, JuliaBox would not do because of Gurobi license)

mlubin:
That's great, but adding a unit test to the package is your insurance that someone (or a gurobi release) won't unintentionally break this code in the future.

iaravena:
Added (test_get_strarray.jl and test_get_strarray.mps), should I  create an independent pull request?

mlubin:
@iaravena, it's better to add to this PR if you can. Just push a new commit to this branch.

test/test_get_strarray.jl
```diff
@@ -0,0 +1,21 @@
...
+# test functions
+@test Gurobi.get_strattrarray(m, "VarName", 1, Gurobi.num_vars(m))[1] == "XONE"
+@test Gurobi.get_strattrarray(m, "ConstrName", 1, Gurobi.num_constrs(m))[1] =="LIM1"
+
+# get variable and constraint names
+varnames = Gurobi.get_strattrarray(m, "VarName", 1, Gurobi.num_vars(m))
+connames = Gurobi.get_strattrarray(m, "ConstrName", 1, Gurobi.num_constrs(m))
+
+# print variable and constraint names
+println(varnames)		# should output ["XONE","YTWO","ZTHREE"]
```
mlubin: This should be a ``@test``, not a printout.

mlubin:
The new test file should be added to the list in ``runtests.jl`` so that it gets run with ``Pkg.test("Gurobi")``.

mlubin:
Confirmed tests passing locally, thanks!

iaravena:
Perfect! thanks for your patience :)

PR #89, opened 2017-03-14
Add delete variables function

iaravena:


src/grb_vars.jl
```diff
@@ -106,3 +106,16 @@ add_bvars!(model::Model, c::Vector) = add_vars!(model, GRB_BINARY, c, 0, 1)
 add_ivars!(model::Model, c::Vector, lb::Bounds, ub::Bounds) = add_vars!(model, GRB_INTEGER, c, lb, ub)
 add_ivars!(model::Model, c::Vector) = add_ivars!(model, GRB_INTEGER, c, -Inf, Inf) 
 
+del_vars!{T<:Real}(model::Model, idx::T) = del_vars!(model, Cint[idx])
```
joehuchette: Don't need to change, but this can be written as
```jl
del_vars!(model::Model, idx::Real) = del_vars!(model, Cint[idx])
```

joehuchette:
Looks good to me.

src/grb_vars.jl
iaravena: I'm trying to be consistent with the code for del_constrs

Issue #94, opened 2017-03-23
Large coefficients


odow:
As raised [here](https://github.com/JuliaOpt/Gurobi.jl/issues/60#issuecomment-288571797), when setting the objective Gurobi silently rounds numbers greater than 1e100 down to 1e100.

```julia
using JuMP, Gurobi
m = Model(solver=GurobiSolver())
@variable(m, x >= 1)
@objective(m, Min, 1.1e100 * x)
solve(m) # :Optimal
getobjectivevalue(m) # 1.0e100 not 1.1e100
```
This also happens when setting other values such as the RHS. This is probably because Gurobi just rounds inputs to GRB_INFINITY (1e100).

Things also get weird around variable bounds. It ignores variable bounds larger than or equal to 1e20 (when +ve and upper, or -ve and lower)
```julia
using JuMP, Gurobi
m = Model(solver=GurobiSolver())
@variable(m, x <= 1e20)
@objective(m, Min, x)
solve(m) # unbounded
# ... or ...
# @variable(m, x >= -1e20)
# @objective(m, Max, x)
```

But when -ve and upper or +ve and lower, it allows up to 1e30
```julia
using JuMP, Gurobi
m = Model(solver=GurobiSolver())
@variable(m, x >= 1e30)
@objective(m, Min, x)
solve(m) # infeasible (default upper bound is 1e30)
# ... or
@variable(m, x <= -1e30)
@objective(m, Max x)
```

These "features" are also present in the Python interface.

Not really sure what we should do about it.

odow:
cc @mburq

odow:
Gurobi confirm that this is expected behaviour:
https://groups.google.com/forum/#!topic/gurobi/M_Bzc6RWMhU

blegat:
It seems that we should do a warning from the Julia wrapper. It might be very helpful and I don't think that it would give a noticeable performance overhead.

PR #96, opened 2017-03-27
Add warnings when using large coefficients

odow:
As [stated by Gurobi](https://groups.google.com/forum/#!topic/gurobi/M_Bzc6RWMhU), "everything that is larger than 1e+100 is silently converted into 1e+100. For bounds, the value is actually 1e+30. So, every bound value that is larger than 1e+30 (smaller than -1e+30) is truncated to 1e+30 (-1e+30)."

This pull requests introduces a warning rather than silently failing. It is made slightly more complicated by the fact we have to propagate the truncation through the `m.obj` vector we keep around to verify Gurobi stores the correct objective.

Closes https://github.com/JuliaOpt/Gurobi.jl/issues/94.

### DIFF
--- a/src/GurobiSolverInterface.jl
+++ b/src/GurobiSolverInterface.jl
@@ -160,16 +160,11 @@ function loadproblem!(m::GurobiMathProgModel, A, collb, colub, obj, rowlb, rowub
   update_model!(m.inner)
   setsense!(m,sense)
 end
-function _truncateobj!(obj::Vector)
-    for i in eachindex(obj)
-        if obj[i] > GRB_INFINITY
-            obj[i] = GRB_INFINITY
-        elseif obj[i] < -GRB_INFINITY
-            obj[i] = -GRB_INFINITY
-        end
-    end
-end
+
 function _truncateobj(v::Real)
+    # Gurobi truncates objective coefficients to +/-GRB_INFINITY
+    # We need to do this to the m.obj vector for consistency
+    #   see #94
     if v > GRB_INFINITY
         return GRB_INFINITY
     elseif v < -GRB_INFINITY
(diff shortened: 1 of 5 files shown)


odow:
cc @blegat @mburq

src/GurobiSolverInterface.jl
```diff
@@ -104,6 +104,7 @@ end
 function loadproblem!(m::GurobiMathProgModel, filename::AbstractString)
     read_model(m.inner, filename)
     m.obj = getobj(m)
+    _truncateobj!(m.obj)
```
blegat: Why don't you put any warnings here ?

src/GurobiSolverInterface.jl
```diff
@@ -155,10 +156,24 @@ function loadproblem!(m::GurobiMathProgModel, A, collb, colub, obj, rowlb, rowub
 
   m.lb, m.ub = rowlb, rowub
   m.obj = copy(obj)
+  _truncateobj!(m.obj)
```
blegat: or here

src/GurobiSolverInterface.jl
```diff
@@ -201,14 +216,18 @@ setconstrLB!(m::GurobiMathProgModel, lb) = (m.changed_constr_bounds = true; m.lb
 setconstrUB!(m::GurobiMathProgModel, ub) = (m.changed_constr_bounds = true; m.ub = copy(ub))
 
 getobj(m::GurobiMathProgModel)     = get_dblattrarray( m.inner, "Obj", 1, num_vars(m.inner)   )
-setobj!(m::GurobiMathProgModel, c) = (m.obj=copy(c); set_dblattrarray!(m.inner, "Obj", 1, num_vars(m.inner), c))
+function setobj!(m::GurobiMathProgModel, c)
+    m.obj = copy(c)
+    _truncateobj!(m.obj)
```
blegat: or here

src/GurobiSolverInterface.jl
odow: should get called in the `add_cvars!` function.

mlubin:
This now has a conflict

odow:
@mlubin resolved.

PR #98, opened 2017-04-12
delete trailing period which breaks issue URL

rdeits:
The error which asks users to comment on #60 has a period immediately after the url, which results in the url being parsed as `https://github.com/JuliaOpt/Gurobi.jl/issues/60.` when displayed in, for example, IJulia. This makes it look like the issue has been removed, and might be preventing users from reporting occurrence of the bug.


joehuchette:
Seems reasonable

PR #99, opened 2017-04-12
Change to objective bug error to warning and only on <v7

odow:
Ref https://github.com/JuliaOpt/Gurobi.jl/issues/60#issuecomment-293706516


odow:
@mlubin thoughts? Or should we just remove this entirely.

mlubin:
Okay with me as is

PR #104, opened 2017-07-13
Fix tests for Gurobi v7.5

joehuchette:


test/mip_01.jl
```diff
@@ -1,6 +1,6 @@
 # an example on mixed integer programming
 #
-#   maximize x + 2 y + 5 z
+#   maximize x + 3 y + 5 z
```
mlubin: What was the issue with the old objective?

joehuchette: It didn't have a unique optimal solution, which is why the test started failing on a new version

PR #106, opened 2017-09-30
MOI compatibility

joaquimg:
Needed updates for MathOptInterfaceGurobi.jl

### DIFF
--- a/src/grb_attrs.jl
+++ b/src/grb_attrs.jl
@@ -96,7 +96,7 @@ function get_intattrarray!(r::Array{Cint}, model::Model, name::String, start::In
     r
 end
 
-function get_intattrlist!(r::Array{Cint}, model::Model, name::String, inds::Vector{I}) where {I<:Integer}
+function get_intattrlist!{I<:Integer}(r::Array{Cint}, model::Model, name::String, inds::Vector{I})
     @assert isascii(name)
     @assert length(r) == length(inds)
     ret = @grb_ccall(getintattrlist, Cint,
@@ -108,7 +108,7 @@ function get_intattrlist!(r::Array{Cint}, model::Model, name::String, inds::Vect
     nothing
 end
 
-function get_intattrlist(model::Model, name::String, inds::Vector{I}) where {I<:Integer}
+function get_intattrlist{I<:Integer}(model::Model, name::String, inds::Vector{I}) 
     r = Array{Cint}(length(inds))
     get_intattrlist!(r::Array{Cint}, model::Model, name::String, inds)
     return r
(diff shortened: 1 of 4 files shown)


src/grb_attrs.jl
```diff
@@ -96,6 +96,24 @@ function get_intattrarray!(r::Array{Cint}, model::Model, name::String, start::In
     r
 end
 
+function get_intattrlist!(r::Array{Cint}, model::Model, name::String, inds::Vector{I}) where {I<:Integer}
```
mlubin: Either drop support for 0.5 or avoid the `where` syntax

mlubin: Probably better to avoid the `where` syntax and make a separate PR later to update this everywhere in the package

src/grb_constrs.jl
```diff
@@ -214,6 +232,75 @@ function add_sos!(model::Model, sostype::Symbol, idx::Vector{Int}, weight::Vecto
         throw(GurobiError(model.env, ret))
     end
 end
+function del_sos!(model::Model, idx::Vector{Cint})
+    # int XPRS_CC XPRSdelsets(XPRSprob prob, int ndelsets, const int mindex[])
```
mlubin: Out of place comment

src/grb_constrs.jl
```diff
@@ -214,6 +232,75 @@ function add_sos!(model::Model, sostype::Symbol, idx::Vector{Int}, weight::Vecto
...
+                     Ptr{Cint}),
+                     model, convert(Cint,numdel), ivec(idx-1) )
+    if ret != 0
+        throw(GurobiError(model.env, ret))
+    end
+end
+
+get_sos_matrix(model::Model) = get_sos(model::Model, 1, num_sos(model))
+function get_sos(model::Model, start::Integer, len::Integer)
+    # int GRBgetsos ( GRBmodel *model,int	*nummembersP,int *sostype,int *beg,int	*ind,doule *weight,int start,int len )
```
mlubin: Also seems out of place, since we don't have comments like these in other wrappers

src/grb_constrs.jl
```diff
@@ -248,3 +335,22 @@ function chg_coeffs!(model::Model, cidx::Vector{Cint}, vidx::Vector{Cint}, val::
         throw(GurobiError(model.env, ret))
     end
 end
+
+function getcoeff!(val::FVec, model::Model, cidx::Integer, vidx::Integer)
+    # int GRBgetcoeff ( 	GRBmodel *model, int constrind, int 	varind, double 	*valP )
```
mlubin: Ditto

src/grb_solve.jl
```diff
@@ -150,3 +158,74 @@ function get_basis(model::Model)
     end
     return cbasis, rbasis
 end
+
+function loadbasis(model::Model, x::Vector)#, status::Symbol = :unstarted, isnew::Vector{Bool} = [false])
```
mlubin: This isn't a direct wrapper so it could use some documentation

src/grb_attrs.jl
blegat: [FemToCleaner](https://github.com/JuliaComputing/FemtoCleaner.jl) can take care of that

PR #109, opened 2017-11-05
make it possible to silence gurobi environment construction

rdeits:
Gurobi 7.5 introduced an academic license notification which prints on every single environment construction (see [this post](https://groups.google.com/forum/#!searchin/gurobi/Academic$20license$20for$20non-commercial$20use|sort:date/gurobi/O7Kht25-OIA/ipRHU_cyAQAJ)). Since my stdout is now filled with Gurobi messages, I figured it would be nice to be able to turn them off. 

This PR just adds a `silent` flag to the `Env()` constructor that temporarily redirects stdout while constructing the environment. 

I also added a test to at least verify that the flag does not cause an error when set to `true`. My initial attempts at actually testing that nothing is printed led down a dark rabbit hole of libuv streams which did not seem worth the effort. If you *do* want a test of the behavior, then I'd suggest making https://github.com/Ismael-VC/Suppressor.jl a test dependency and using its `@capture_stdout`. Just let me know if you want that.

### DIFF
--- a/src/grb_env.jl
+++ b/src/grb_env.jl
@@ -3,11 +3,40 @@
 
 type Env
     ptr_env::Ptr{Void}
-    
-    function Env()
+
+    """
+        Env(silent::Bool=false)
+
+    Construct a Gurobi environment. The environment maintains parameters of the
+    Gurobi solver and its license. In particular, constructing an environment
+    will obtain a new Gurobi license token. If you are operating in a setting
+    where the number of concurrent Gurobi tokens is limited, you may want to
+    construct a single environment ahead of time and share it among all the
+    models solved by your program.
+
+    The `silent` option temporarily redirects the standard output stream
+    while constructing the environment. This may be useful in preventing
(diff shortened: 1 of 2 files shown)


mlubin:
This could be viewed as a means to circumvent Gurobi's licensing restrictions. Why not just create an environment once and use it for all your Gurobi models? It's the `env=` parameter to `GurobiMathProgModel`:
https://github.com/JuliaOpt/Gurobi.jl/blob/eb12928e5b51808a4916f3c9963d8ccd1f8a5c11/src/GurobiSolverInterface.jl#L21

rdeits:
Yeah, I understand that concern. I'm already passing around the `env` in a lot of places, but it is kind of annoying to have to wire a single global gurobi environment through all of my various functions. But, that said, I can always just do the stdout redirection in my own code.

mlubin:
Complain about the printout to support@gurobi.com or on the thread you linked.

PR #121, opened 2018-03-29
Advance routines

Thuener:
Add routines of Gurobi to get information from tableau

### DIFF
--- a/src/grb_advroutines.jl
+++ b/src/grb_advroutines.jl
@@ -1,26 +1,43 @@
-type GRBsvec
+mutable struct GRBsvec
     len::Cint
     ind::Ptr{Cint}
     val::Ptr{Cdouble}
 end
 
-function get_tableaurow(model::Model, cidx::Integer)
-    len = num_vars(model)+num_constrs(model)
-    grb_v = GRBsvec(0, pointer(Array{Cint}(len)) , pointer(Array{Cdouble}(len) ))
+function get_tableaurow_grb(model::Model, cidx::Integer)
+	len = num_vars(model)+num_constrs(model)
+	idx = Array{Cint}(len)
+	val = Array{Cdouble}(len)
+
+    grb_v = GRBsvec(0, pointer(idx) , pointer(val))
     ret = @grb_ccall(BinvRowi, Cint, (
         Ptr{Void},
         Cint,
-        Ptr{GRBsvec}),
-        model, convert(Cint, cidx-1), &grb_v)
+        Ref{GRBsvec}),
+        model, convert(Cint, cidx-1), grb_v)
     if ret != 0
         throw(GurobiError(model.env, ret))
     end
+	return idx[1:grb_v.len]+1, val[1:grb_v.len]
+end
 
-    idx = unsafe_wrap(Array, grb_v.ind, grb_v.len)
-    val = unsafe_wrap(Array, grb_v.val, grb_v.len)
-    spv = spzeros(len)
-    spv[idx+1] = val
-    return spv
+function get_tableaurow!(row::Vector{Float64}, model::Model, cidx::Integer)
+	idx, val = get_tableaurow_grb(model, cidx)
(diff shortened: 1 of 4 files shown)


src/grb_advroutines.jl
```diff
@@ -0,0 +1,53 @@
+mutable struct GRBsvec
+    len::Cint
+    ind::Ptr{Cint}
+    val::Ptr{Cdouble}
+end
+
+function get_tableaurow_grb(model::Model, cidx::Integer)
+	len = num_vars(model)+num_constrs(model)
+	idx = Array{Cint}(len)
+	val = Array{Cdouble}(len)
```
odow: Looks like the indentation is off?

Thuener: What is the default here? Spaces or tabs?

joehuchette: 4 spaces

src/grb_advroutines.jl
```diff
@@ -0,0 +1,53 @@
+mutable struct GRBsvec
+    len::Cint
+    ind::Ptr{Cint}
+    val::Ptr{Cdouble}
+end
+
+function get_tableaurow_grb(model::Model, cidx::Integer)
```
mlubin: It's best for this function to have a name that matches the function being wrapped. See e.g., https://github.com/JuliaOpt/Gurobi.jl/blob/93b4d3e37c3d8ed2ec37b59d207504537e85e173/src/grb_quad.jl#L4.

Thuener: BinvRowi is a strange name for a function in Julia. The other problem is that this name is misleading, this is not actually the row from B^-1 is the row from A B^-1. http://www.gurobi.com/documentation/7.5/refman/c_grbbinvrowi.html
Should I call it ABinv_row?

odow: How about `c_api_binvrowi`

src/grb_advroutines.jl
```diff
@@ -0,0 +1,53 @@
...
+        throw(GurobiError(model.env, ret))
+    end
+    return idx[1:grb_v.len]+1, val[1:grb_v.len]
+end
+
+function get_tableaurow!(row::Vector{Float64}, model::Model, cidx::Integer)
+    idx, val = get_tableaurow_grb(model, cidx)
+    row[idx] = val
+    return nothing
+end
```
odow: I'd prefer not to have these helper functions in Gurobi.jl. They don't get tested and lead to multiple ways to do the same thing.

src/grb_advroutines.jl
```diff
@@ -0,0 +1,53 @@
...
+
+function get_tableaurow!(BinvA::Matrix{Float64}, model::Model, cidx::Integer)
+    idx, val = get_tableaurow_grb(model, cidx)
+    BinvA[cidx, idx] = val
+    return nothing
+end
```
odow: Ditto

src/grb_advroutines.jl
```diff
@@ -0,0 +1,53 @@
...
+
+function get_tableaurow(model::Model, cidx::Integer)
+    len = num_vars(model)+num_constrs(model)
+    spv = zeros(len)
+    get_tableaurow!(spv, model, cidx)
+    return spv
+end
```
odow: Ditto

src/grb_advroutines.jl
```diff
@@ -0,0 +1,53 @@
...
+
+function get_basisidx(model::Model)
```
odow: `c_api_getbasishead`

src/grb_advroutines.jl
Thuener: c_api_abinvrowi?

src/grb_advroutines.jl
Thuener: Do you prefer to remove get_tableaurow! ?

PR #127, opened 2018-05-01
exposing GRBread()

antonhinneck:
Hi, 

I created this PR based on a [post of me on the Julia forum](https://discourse.julialang.org/t/gurobi-mip-start-calling-grbread-to-provide-starting-vector-using-a-mst-sol-file/10610).

Gurobi has a built in function to read *.sol / *.mst to automatically construct starting vectors for mip starts, which is called `GRBread`.

I just dropped the "GRB" that is at the beginning of the original functions in Gurobi. However, "read" must be quite common as a function name ...  "grb_read" would probably be more unique.

Anton


odow:
Thanks Anton!

Can you add a test please? It would basically just need to do
- make a model
- save the sol file
- read in the sol file
- check the mip start
- delete the files

If possible, use only the Gurobi API and don't use JuMP or MathProgBase :)

antonhinneck:
I added a test file, only using the Gurobi API.

I, again, can't get gurobi to indicate that the mip start was used.
To be honest, I didn't really get why `simple_model.colVal .= NaN` was making a difference, to begin with :)
I verified though, that the correct vector is present in the mod and will update the test file once a find a solution :)

antonhinneck:
Hi Oscar, everything works now. The problem were both, the Presolve and Heuristics parameter in Gurobi. If the presolve or any heuristic produces a good enough solution right at the start, the mip start is not used. Setting both these parameters to 0 resolves the issue.
I updated the test file.

test/test_GRBread.jl
```diff
@@ -0,0 +1,83 @@
...
+optimize(simple_model)
+println("-----------------------------")
+println()
+
+## WRITE SOLUTION FILES
+#----------------------
+
+write_model(simple_model, "simple_out.sol")
+write_model(simple_model, "simple_out.mst")
+
```
odow:
```julia
@test isfile("simple_out.sol")
@test isfile("simple_out.mst")
```

test/test_GRBread.jl
```diff
@@ -0,0 +1,83 @@
...
+add_constr!(simple_model, [0.5, 2., 1., 0.], '<', 6.)
+
+setparam!(simple_model, "Heuristics", 0.0)
+setparam!(simple_model, "Presolve", 0)
+
+update_model!(simple_model)
+
+optimize(simple_model)
+println("-----------------------------")
+println()
```
odow: You can remove all print lines

test/test_GRBread.jl
```diff
@@ -0,0 +1,83 @@
+using Gurobi
+
+dir = pwd()
+cd(dir)
```
odow: you don't need these lines

test/test_GRBread.jl
```diff
@@ -0,0 +1,83 @@
...
+## WRITE SOLUTION FILES
+#----------------------
+
+write_model(simple_model, "simple_out.sol")
+write_model(simple_model, "simple_out.mst")
+
+## READ SOLUTION FILE
+#--------------------
+
+start_1 = Gurobi.get_dblattrarray(simple_model, "Start", 1, 4)
```
odow: `@test start_1 == [1.0e101, 1.0e101, 1.0e101, 1.0e101]` (n.b. I didn't test this)

test/test_GRBread.jl
```diff
@@ -0,0 +1,83 @@
...
+println("Current MIP start vector:")
+println(start_1)
+println()
+
+println("----------------------")
+println("Reading solution file.")
+println("----------------------")
+println()
```
odow: ditto with the prints

test/test_GRBread.jl
```diff
@@ -0,0 +1,83 @@
...
+Gurobi.read(simple_model, "simple_out.sol")
+update_model!(simple_model)
+
+println("Updated MIP start vector:")
+start_2 = Gurobi.get_dblattrarray(simple_model, "Start", 1, 4)
```
odow: `@test start_2 == [12.0, 0.0, 0.0, 24.0]`

test/test_GRBread.jl
```diff
@@ -0,0 +1,83 @@
...
+println(start_2)
+
+## VERIFY MIP START
+#------------------
+
+#The parameters Presolve and Heuristics must be set to 0
+println()
+println("-----------------------------")
+optimize(simple_model)
```
odow: You don't need this

test/test_GRBread.jl
```diff
@@ -0,0 +1,83 @@
...
+add_constr!(simple_model, [2., 5., 4., -1.], '>', 0.)
+
+add_constr!(simple_model, [0.5, 2., 1., 0.], '<', 6.)
+
+setparam!(simple_model, "Heuristics", 0.0)
+setparam!(simple_model, "Presolve", 0)
+
+update_model!(simple_model)
+
+optimize(simple_model)
```
odow: Add a test for the objective value or something to confirm it solved correctly?

antonhinneck:
Hi Oscar, 

- I added my file "test_read" to runtests.jl,
- wrapped everything in `@testset`,
- added `@test` checks,
- and confirmed that `Pkg.test("Gurobi")` produces `INFO: Gurobi tests passed` as a result.

Also, 
- I appended `Base.read` to the `import` section,
- changed my simple model (I realized that the model, if it was an LP, would have infinite solutions. Being a MIP it still had 24 solutions with an optimal objective. I made sure it only has one optimal solution. The optimal objective (_36.0_) is also checked for during the test.),
- and removed all the `println()` calls.

Thanks for your help, I appreciate it a lot :)

odow:
Thanks @antonhinneck!

PR #131, opened 2018-06-24
use Ref instead of Array when getting Gurobi parameters

rdeits:
Allocating an entire `Array` is quite expensive and not actually necessary just to be able to pass a reference to a C function. Using a `Ref` saves 80 bytes of memory on Julia v0.6.3, and I think Julia v0.7 should be able to eliminate the memory allocation entirely. 

Before: 

```julia
julia> @btime Gurobi.get_intattr_old(m, $("NumQNZs"))
  80.248 ns (1 allocation: 96 bytes)
0
```

After: 

```julia
julia> @btime Gurobi.get_intattr(m, $("NumQNZs"))
  70.103 ns (1 allocation: 16 bytes)
0
```

### DIFF
--- a/src/grb_attrs.jl
+++ b/src/grb_attrs.jl
@@ -8,79 +8,79 @@
 
 function get_intattr(model::Model, name::String)
     @assert isascii(name)
-    a = Array{Cint}(1)
+    a = Ref{Cint}()
     ret = @grb_ccall(getintattr, Cint,
-        (Ptr{Void}, Ptr{UInt8}, Ptr{Cint}),
+        (Ptr{Void}, Ptr{UInt8}, Ref{Cint}),
         model, name, a);
     if ret != 0
         throw(GurobiError(model.env, ret))
     end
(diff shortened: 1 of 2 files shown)


mlubin:
Thanks! This code was almost certainly written before `Ref` existed.

rdeits:
Wow, cool :slightly_smiling_face: 

I'm working on a few more performance improvements--Ideally, I'd like to be able to modify a Gurobi problem in-place without necessarily allocating memory.

mlubin:
> One thing I'm considering is a possibly crazy idea to do the conversion in-place (since we can fit two Int32s inside an Int64) and then putting them back at the end: 

What about storing an extra `Vector{Cint}` or two in the model object and copying the coefficients over? They can be resized on occasion if needed, still eliminating most allocations if there are lots of modifications.

rdeits:
That's also a good idea and substantially less crazy than what I was proposing.

PR #137, opened 2018-07-25
concretely type the inner model

rdeits:
Requires https://github.com/JuliaOpt/LinQuadOptInterface.jl/pull/39


odow:
We probably need to tag a new LQOI release.

odow:
Even though this will break master (because of the LQOI bound in REQUIRE), I'm going to merge this because there are a few PR's backing up and I want to test them all in anticipation of the upcoming tags of MOI and LQOI.

PR #139, opened 2018-07-28
Update tests to reflect changes in MOIT

odow:
- [x] Requires https://github.com/JuliaOpt/MathOptInterface.jl/pull/434
- [ ] Requires new tag of LQOI with bumped MOI


odow:
Close in favor of #146

PR #141, opened 2018-07-31
return primal feasible when possible, even when user limits are reached

### LINKED ISSUE #140: `NaN` solution when MIP solution hits the time limit (via LinQuadOptInterface)
When an MIP hits its time limit before reaching optimality, we return `MOI.UnknownResultStatus` for both the primal and dual status of the problem. However, it's often the case that at least one feasible solution has been found before the time limit, but the user can't retrieve this solution because LQOI sees the primal UnknownResultStatus and doesn't extract the primal solution. Instead, when we reach the time limit, I think we should ask Gurobi if it has a feasible solution and return `FeasiblePoint` for the primal status if so. 

Does that seem reasonable?

rdeits:
fixes #140

### DIFF
--- a/test/MOIWrapper.jl
+++ b/test/MOIWrapper.jl
@@ -1,8 +1,7 @@
-using Gurobi, Base.Test, MathOptInterface, MathOptInterface.Test, LinQuadOptInterface
+using Gurobi, Base.Test, MathOptInterface, MathOptInterface.Test
 
 const MOI  = MathOptInterface
 const MOIT = MathOptInterface.Test
-const LQOI = LinQuadOptInterface
 
 @testset "Unit Tests" begin
     config = MOIT.TestConfig()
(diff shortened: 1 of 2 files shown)


rdeits:
It's been working for me today. I'm not sure what the right test case for this would be, though. We'd need a MIP that Gurobi definitely can't solve to optimality within some given time limit, but for which we can provide a feasible warm-start.

odow:
Turn Persolve off. Give warmstart. Set time limit to 0.0 or something v small?

rdeits:
Ok, done. I'm using the solution limit instead of the time limit because I suspect it will be less flaky.

test/MOIWrapper.jl
```diff
@@ -189,3 +190,42 @@ end
...
+    item_weights = rand(N)
+    item_values = rand(N)
+    MOI.addconstraint!(m,
+        MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(item_weights, x), 0.0),
+        MOI.LessThan(10.0))
+    MOI.set!(m, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}(),
+        MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(.-item_values, x), 0.0))
+    MOI.optimize!(m)
+
+    @test LQOI.get_termination_status(m) == MOI.SolutionLimit
```
odow: You should test MOI status here. These LQOI methods shouldn't be queried by the user.

rdeits: done

rdeits:
Rebased on master and tested locally: 

```
Test Summary:      | Pass  Total
MathProgBase Tests |  240    240
Warning: Constant in scalar function moved into set.
Warning: Constant in scalar function moved into set.
Test Summary:          | Pass  Total
MathOptInterface Tests | 2447   2447
Test Summary:         | Pass  Total
changing coefficients |   17     17
Test Summary:                  | Pass  Total
Empty constraints (Issue #142) |    2      2
INFO: Gurobi tests passed
```

odow:
Thanks @rdeits

PR #145, opened 2018-07-31
Pass constant objective terms all the way to the solver

### LINKED ISSUE #111: GurobiSolver ignores constant term in objective
It appears that `GurobiSolver` ignores any constant terms in the JuMP model's cost function. For example:

```julia
using JuMP, Gurobi

model = Model(solver=GurobiSolver(OutputFlag=1))
@variable model x
@objective model Min x^2 - 1
solve(model)
```

The gurobi output shows:

```
Optimize a model with 0 rows, 1 columns and 0 nonzeros
Model has 1 quadratic objective term
Coefficient statistics:
  Matrix range     [0e+00, 0e+00]
  Objective range  [0e+00, 0e+00]
  QObjective range [2e+00, 2e+00]
  RHS range        [0e+00, 0e+00]
Presolve removed 0 rows and 1 columns
Presolve time: 0.00s
Presolve: All rows and columns removed

Barrier solved model in 0 iterations and 0.00 seconds
Optimal objective 0.00000000e+00
```

but the optimal objective value is actually `-1`, as confirmed by JuMP:

```julia
getvalue(getobjective(model))
# -1.0
```

Of course, for most optimization problems this is fine. Any constant term in the objective doesn't affect the minimizer, and JuMP still reports the correct final objective with `getvalue()`. 

However, there is one case in which this absolutely does matter: mixed-integer programming. When solving an MIP, Gurobi checks optimality by measuring (among other things), the relative gap between the best incumbent and the best bound. Since this gap is relative to the actual cost value, adding an arbitrary constant offset (or failing to add one) will affect whether the MIP gap is considered small enough. 

This is causing some trouble for me, since I see very different solve times depending on how the particular initial state of my system affects the (eliminated) constant term in the cost. 

Is there any way to turn this on in Gurobi.jl? I'd be happy to do the work myself, but I'd appreciate some pointers on where to look, since I haven't used the internal Gurobi interface at all.

rdeits:
Fixes #111 (for the MOI interface). Requires https://github.com/JuliaOpt/LinQuadOptInterface.jl/pull/44

This currently contains a really gross workaround for the issues discussed in https://github.com/JuliaOpt/LinQuadOptInterface.jl/pull/44#issuecomment-409373755 . I'm happy to wait until we have a better solution for that before merging; I just wanted to get this PR opened so I won't forget about the changes.

### DIFF
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -251,6 +251,16 @@ function LQOI.set_linear_objective!(instance::GurobiOptimizer, columns::Vector{I
     update_model!(instance.inner)
 end
 
+function LQOI.set_constant_objective!(instance::GurobiOptimizer, value::Real)
+    set_dblattr!(instance.inner, "ObjCon", value)
+    if num_vars(instance.inner) > 0
+        # Work-around for https://github.com/JuliaOpt/LinQuadOptInterface.jl/pull/44#issuecomment-409373755
+        set_dblattrarray!(instance.inner, "Obj", 1, 1,
+            get_dblattrarray(instance.inner, "Obj", 1, 1))
+    end
+    update_model!(instance.inner)
+end
+
 function LQOI.change_objective_sense!(instance::GurobiOptimizer, symbol)
     if symbol == :min
         set_sense!(instance.inner, :minimize)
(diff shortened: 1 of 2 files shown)


rdeits:
x-ref: https://groups.google.com/forum/#!topic/gurobi/p2uJgx8Hn7Q

odow:
Even though this will break master (because of the LQOI bound in REQUIRE), I'm going to merge this because there are a few PR's backing up and I want to test them all in anticipation of the upcoming tags of MOI and LQOI.

PR #150, opened 2018-08-22
Some more fixes for Julia 1.0

daschw:


deps/build.jl
```diff
@@ -1,4 +1,7 @@
 using Compat
+@static if VERSION >= v"0.7"
+  using Libdl
```
blegat: Need 4-spaces indent

odow:
Can you confirm that the tests pass, and that these are the only imports required from MathProgBase?

daschw:
No, I have to update and add some more MathProgBase imports. Thanks for the hint and sorry for my sloppiness! I can update the PR later today.

blegat:
@daschw The cleanest way is to do what is done in https://github.com/JuliaOpt/Ipopt.jl/pull/127, i.e. defining an `MPB` constant and explicitely prefixing with `MPB.`

ExpandingMan:
For me (old version of Gurobi) on line 351 you need to qualify `getobj` as `MPB.getobj` or else it throws an undefined error.

daschw:
Thanks a lot @ExpandingMan, I missed that on 1.0.
Now the tests pass locally on 0.6.4 and on 1.0.

blegat:
@ExpandingMan is it compile time or run time ? What happens in the second run ? We have similar issues with https://github.com/JuliaPolyhedra/Polyhedra.jl/issues/110

ExpandingMan:
It definitely appears to be run-time as it happens in the middle of the execution of a function, though of course in Julia I'm never 100% certain that nothing is compiling.

blegat:
If you execute this function twice, you know that the second time, nothing is compiling.

ExpandingMan:
Nope, it's definitely all run-time, compile time seems just fine.

That's actually a huge relief because that means it fixable.

rdeits:
Can you run the profiler and get a sense of what part is slow?

ExpandingMan:
What profiler do you use?  I'm not aware of any good ones.  I do know the slowness occurs in `build`.

blegat:
You can use https://github.com/timholy/ProfileView.jl

ExpandingMan:
That package doesn't look maintained and surely will not work on 1.0.

The regression only seems to be on 1.0, and probably only with `MathProgBase`, and before this PR `Gurobi` was not working on 1.0 at all.  Therefore I suggest that you merge this PR now and we can look into the regressions.  This will definitely be a priority for me at some point, so I'll look into it as soon as I can, but please don't let that stop you from merging this.

blegat:
Yes, the issue should be moved to another thread. Note that ProfileView is maintainted and is being update to Julia v1.0: https://github.com/timholy/ProfileView.jl/pull/96

rdeits:
@ExpandingMan you can also store a profile using the built-in Profile and then load it with ProfileView on 0.6: https://github.com/timholy/ProfileView.jl#saving-profile-data-manually

Juno also has a built-in profile viewer: http://docs.junolab.org/latest/man/juno_frontend.html#Profiler-1

PR #153, opened 2018-08-29
fix #152

### LINKED ISSUE #152: large performance regression in calling `JuMP.build` using `MathProgBase` in 1.0
In my case from a few seconds to almost 2 minutes.  I know it's coming from the call to `JuMP.build` on the model, but I don't know where yet.

Sorry for the lack of details, will look into it as soon as I get the chance, wanted to open this issue to keep track.

ExpandingMan:
Ok, here is a fix to a rather scary regression that I was afraid would be a nightmare to fix.  The problem is [this](https://github.com/JuliaLang/julia/issues/28948) Julia "bug".  It's really just that `sparse` is missing the appropriate methods for computing a sparse matrix from the transpose of a sparse matrix.  Fortunately, those methods are only missing from `sparse`, calling the `SparseMatrixCSC` constructor works just fine.

I have tested this on my large problems and it now seems at least as fast on 1.0 as it was on 0.6.4.

I believe this would have affected both `MPB` and `MOI`.

### DIFF
--- a/src/grb_constrs.jl
+++ b/src/grb_constrs.jl
@@ -1,4 +1,4 @@
-# work around for julia issue #28948
+# work around for julia issue #28948, Gurobi.jl issue #152
 if VERSION ≤ v"0.7-"
     sparse_transpose(A) = sparse(transpose(A))
 else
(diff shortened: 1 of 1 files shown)


ExpandingMan:
Fixes #152

PR #161, opened 2018-10-05
Fix broken MPB callback interface

mlubin:
Closes #159 

I'm not able to test this. @chriscoey, @CBongiova please confirm that this fixes the issue.

PR #162, opened 2018-10-06
Compat fix for 1.0

mlubin:
Reported at https://github.com/JuliaOpt/Gurobi.jl/pull/161#issuecomment-427577970.

Untested. Would be nice if someone confirms this works.


CBongiova:
@mlubin Tested. Works for me thanks :)

PR #167, opened 2018-11-06
Style and performance improvements

odow:
Best reviewed commit-wise.

<s>

## Commit 1

The first commit changes the process for changing variable bounds from one that made a few vectors, looped through data <s>twice</s>thrice, and then made two C calls, to a simple loop that sets the bounds element-wise. For small numbers of variables (e.g., changing the bound of a single variable), this PR comes out ahead. For moderate numbers (e.g, 10 variables), it is ~5% slower, and then for large numbers (e.g., > 100) it's back in front. I must have tested on different LQOI commits. Here are the latest benchmarks:

```julia
using Gurobi, BenchmarkTools
const MOI = Gurobi.MOI

function setup(N)
    model = Gurobi.Optimizer()
    x = MOI.add_variables(model, N)
    c = MOI.add_constraints(model, MOI.SingleVariable.(x), Ref(MOI.EqualTo(1.0)))
    return model, c
end

function bench(model, constraints)
    for constraint in constraints
        MOI.set(model, MOI.ConstraintSet(), constraint, MOI.EqualTo(2.0))
    end
end
```

### Before

```julia
julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^0))
2.600 μs (14 allocations: 1.08 KiB)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^1))
25.042 μs (140 allocations: 10.78 KiB)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^2))
271.362 μs (1400 allocations: 107.81 KiB)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^3))
6.075 ms (14978 allocations: 1.07 MiB)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^4))
427.902 ms (158978 allocations: 10.82 MiB)
```

### After

```julia
julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^0))
  1.437 μs (6 allocations: 336 bytes)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^1))
  13.958 μs (60 allocations: 3.28 KiB)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^2))
  234.004 μs (600 allocations: 32.81 KiB)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^3))
  5.629 ms (6978 allocations: 343.41 KiB)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^4))
  431.477 ms (78978 allocations: 3.49 MiB)
```
</s>

Things slow down for more than 10^4 variables. We really need to be able to set multiple constraints sets simultaneously...

## Commit 2

Instead of querying the full rays, we just try to access the first element. This results in a 2x improvement. I looked at changing from a try-catch to an explicit C call to get the return value, but it was a pretty minor improvement for extra low-level code lying around.

## Commit 3

No real performance difference, but removes 1 allocation each time we query the status.

## Commit 4

Probably used infrequently, but it seemed pointless to perform a copy here rather than switching from 0-indexing to 1-indexing in-place.

## Commit 5

Style updates.

### DIFF
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -112,28 +112,30 @@ function LQOI.change_variable_bounds!(model::Optimizer,
         set_dblattrlist!(model.inner, "UB", upper_cols, upper_values)
     end
     update_model!(model.inner)
+    return
 end
 
 function LQOI.get_variable_lowerbound(model::Optimizer, column::Int)
-    get_dblattrelement(model.inner, "LB", column)
+    return get_dblattrelement(model.inner, "LB", column)
 end
 
 function LQOI.get_variable_upperbound(model::Optimizer, column::Int)
-    get_dblattrelement(model.inner, "UB", column)
+    return get_dblattrelement(model.inner, "UB", column)
 end
 
 function LQOI.get_number_linear_constraints(model::Optimizer)
-    num_constrs(model.inner)
+    return num_constrs(model.inner)
 end
 
 function LQOI.add_linear_constraints!(model::Optimizer,
         A::LQOI.CSRMatrix{Float64}, sense::Vector{Cchar}, rhs::Vector{Float64})
     add_constrs!(model.inner, A.row_pointers, A.columns, A.coefficients, sense, rhs)
     update_model!(model.inner)
+    return
 end
 
 function LQOI.get_rhs(model::Optimizer, row::Int)
-    get_dblattrelement(model.inner, "RHS", row)
+    return get_dblattrelement(model.inner, "RHS", row)
 end
 
 function LQOI.get_linear_constraint(model::Optimizer, row::Int)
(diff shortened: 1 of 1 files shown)


src/MOIWrapper.jl
```diff
@@ -86,54 +86,40 @@ LQOI.backend_type(::Optimizer, ::MOI.Nonnegatives)         = Cchar('>')
...
-    upper_cols   = fill(0, number_upper_bounds)
-    upper_values = fill(0.0, number_upper_bounds)
-    lower_index = 1
-    upper_index = 1
     for (column, bound, sense) in zip(columns, new_bounds, senses)
         if sense == Cchar('L')
-            lower_cols[lower_index]   = column
-            lower_values[lower_index] = bound
-            lower_index += 1
+            set_dblattrelement!(model.inner, "LB", column, bound)
```
joaquimg: what about: http://www.gurobi.com/documentation/8.0/refman/c_grbsetdblattrlist.html ?

joaquimg:
I have a few questions about point 1. 
2 to 5 look great!

src/MOIWrapper.jl
joaquimg: Got it,
Did you benchmark with 1000 variables?
If the new method is better for 1000 variables I definatly vote for it. (And you can merge). Otherwise I would like more info.

odow: I've added benchmarks to the top post. The new method is actually better for all sizes < 10^4... I must have conducted the previous benchmarks on different LQOI commits or something.

mlubin:
```
julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^2))
  234.004 μs (600 allocations: 32.81 KiB)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^3))
  5.629 ms (6978 allocations: 343.41 KiB)

julia> @btime bench(model, constraints) setup=((model, constraints) = setup(10^4))
  431.477 ms (78978 allocations: 3.49 MiB)
```

Why doesn't the setup time grow linearly with the size of the problem? What are we doing wrong?

src/MOIWrapper.jl
```diff
@@ -448,27 +448,35 @@ function LQOI.get_farkas_dual!(instance::Optimizer, place)
     place .*= -1.0
 end
 
-# TODO(odow): remove try/catch
-function hasdualray(model::Optimizer)
+function has_dual_ray(model::Optimizer)
     try
-        get_dblattrarray(model.inner, "FarkasDual", 1, num_constrs(model.inner))
+        # Note: for performance reasons, we only try getting the first element.
+        Gurobi.get_dblattrarray(model.inner, "FarkasDual", 1, 1)
```
mlubin: What happens for problems with 0 constraints?

odow:
This will teach me for benchmarking the wrong function... I should have timed `set_variable_bounds` explicitly. The majority of the time is in LQOI which gives me a different place to go hunting. Re our email discussion earlier, I'm getting confused with all these layers...

```julia
using Gurobi, BenchmarkTools
const MOI = Gurobi.MOI
const LQOI = Gurobi.LQOI

function constr_setup2(N)
    model = Gurobi.Optimizer()
    x = MOI.add_variables(model, N)
    sets = fill(MOI.EqualTo(1.0), N)
    return model, MOI.SingleVariable.(x), sets
end
function constr_run2(model, variables, sets)
    LQOI.set_variable_bounds(model, variables, sets)
end

@btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^1));
@btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^2));
@btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^3));
@btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^4));
```

With commit 1:
```julia
julia> @btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^1));
  7.636 μs (8 allocations: 1.09 KiB)

julia> @btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^2));
  100.170 μs (8 allocations: 6.00 KiB)

julia> @btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^3));
  901.529 μs (8 allocations: 51.64 KiB)

julia> @btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^4));
  9.004 ms (13 allocations: 508.59 KiB)
```

After reverting commit 1:
```julia
julia> @btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^1));
  2.691 μs (16 allocations: 2.28 KiB)

julia> @btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^2));
  8.211 μs (16 allocations: 12.31 KiB)

julia> @btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^3));
  94.833 μs (16 allocations: 107.39 KiB)

julia> @btime constr_run2(m, v, s) setup = ((m, v, s) = constr_setup2(10^4));
  1.062 ms (29 allocations: 1.03 MiB)
```

odow:
The setting of variable bounds could be improved, but this PR is not the place. I've just dropped the commit. Now this PR just contains minor, in-offensive updates.

PR #175, opened 2018-12-05
Fix ray queries for Gurobi 8

odow:
It seems like something must have changed going from v7.0.x to v7.5.x because Gurobi no longer throws an error querying 0 elements.

### DIFF
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -495,8 +495,9 @@ end
 
 function has_dual_ray(model::Optimizer)
     try
-        # Note: for performance reasons, we only try to get 0 elements.
-        Gurobi.get_dblattrarray(model.inner, "FarkasDual", 1, 0)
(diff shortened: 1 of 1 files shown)


englhardt:
I can confirm this fix -- tests now pass with Gurobi v8.1.0.

src/MOIWrapper.jl
```diff
@@ -495,8 +495,9 @@ end
 
 function has_dual_ray(model::Optimizer)
     try
-        # Note: for performance reasons, we only try to get 0 elements.
-        Gurobi.get_dblattrarray(model.inner, "FarkasDual", 1, 0)
+        # Note: for performance reasons, we try to get 1 element because as of
+        # Gurobi 8, we can query 0 elements without error.
```
blegat: can -> cannot ?

odow: Good catch.

PR #176, opened 2018-12-05
Introduce GUROBI_JL_SKIP_LIB_CHECK environment variable.

tkoolen:
I think this would be useful for CI testing of packages for which a subset of the tests require Gurobi. For such packages, it would be nice to be able to put tests that depend on Gurobi in an`if haskey(ENV, "CI") ... end` block.

On a machine with a Gurobi license, Gurobi needs to be a test dependency, so it should be in the `REQUIRE`/`Project.toml` file. But on CI, testing currently errors because Gurobi cannot be built due to the `!found` check, even though Gurobi would not actually be used at test time. Setting the `GUROBI_JL_SKIP_LIB_CHECK` variable on CI could then be used to make Gurobi installable.

I can add a note to the readme if you're OK with this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,11 @@ Here is the procedure to setup this package:
 
 4. Now, you can start using it.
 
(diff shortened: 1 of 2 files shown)


odow:
This seems reasonable. I've run into this quite a few times as well.

tkoolen:
Added a readme note. My editor removed trailing whitespace in the readme. Let me know if you want me to undo that.

odow:
It's all good; whitespace must die. I'll leave this open for a day to give other people time to chip in if they strongly object to this.

tkoolen:
OK, thanks!

PR #187, opened 2019-01-30
Allow passing Env to Optimizer

odow:
Restores the functionality that existed in MPB: https://github.com/JuliaOpt/Gurobi.jl/pull/57

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -56,7 +56,7 @@ end
 LQOI.LinearQuadraticModel(::Type{Optimizer}, env) = Model(env::Env,"defaultname")
 
 """
-    Optimizer(;kwargs...)
+    Optimizer(env::Gurobi.Env = Gurobi.Env(); kwargs...)
 
 Create a new Optimizer object.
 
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -37,32 +37,46 @@ const SUPPORTED_CONSTRAINTS = [
...
+    free to overide this as follows `Gurobi.Optimizer(InfUndbInfo=0)`.
+    """
+    function Optimizer(env::Union{Nothing, Env} = nothing; kwargs...)
+        model = new()
+        model.env = env
+        model.params = Dict{String, Any}()
+        MOI.empty!(model)
+        for (name, value) in kwargs
+            model.params[string(name)] = value
+            setparam!(model.inner, string(name), value)
```
blegat: What happens if the name is not a valid error ? Does Gurobi throws an helpful error ?
We should maybe add a comment about that

test/MOI_wrapper.jl
```diff
@@ -259,3 +259,39 @@ end
...
+        finalize(model_1.inner)
+        @test !Gurobi.is_valid(model_1.inner.env)
+    end
+    @testset "Env when emptied" begin
+        @testset "User-provided" begin
+            env = Gurobi.Env()
+            model = Gurobi.Optimizer(env)
+            @test model.inner.env === env
+            MOI.empty!(model)
+            @test model.inner.env === env
```
blegat: We could add a test with Gurobi.is_valid

src/MOI_wrapper.jl
odow: `Gurobi.Optimizer(bad_kwarg = 1)` gives `Error: Unrecognized parameter name: bad_kwarg.` So that seems fine for now.

PR #188, opened 2019-01-31
Fix performance issues

### LINKED ISSUE #184: Gurobi is slow with MOI, likely due to high overhead of model updates
Here's code for a small working example. The problem is a network flow problem over a complete graph: `n` points are generated in the unit square, corresponding to vertices of the graph. Vectors `x` and `y` of length `n` and sum `1` correspond to vertex-wide supply and demand. Script:
```
using JuMP, GLPK, Gurobi, Random

# build a network flow JuMP model
# n (dense) constraints, n^2 variables
function build_jumpmodel(x, y, dm, optimizer)
    n = length(x)
	@assert n == length(y)

	m = JuMP.Model(with_optimizer(optimizer))
	@variable(m, f[1:n, 1:n] >= 0)
	@objective(m, Min, sum(dm[i,j] * f[i,j] for i in 1:n, j in 1:n))


	# correct net change at each voxel to transform x to y
	for i in 1:n
		@constraint(m,
            sum(f[j,i] for j in 1:n)
            - sum(f[i,j] for j in 1:n)
            == y[i] - x[i]
        )
	end
    
    return m
end

function random_network_flow(n, optimizer)
    Random.seed!(123)
    # n random vertices in the unit square
    # random weight vectors, each summing to 1
    vert_locs = rand(2,n)
    dm = zeros(n,n)
    for i in 1:n, j in 1:n
        dm[i,j] = (vert_locs[:,i] - vert_locs[:,j]).^2 |> sum |> sqrt
    end
    x = rand(n); x ./= sum(x)
    y = rand(n); y ./= sum(y)
    
    build_time = @elapsed m = build_jumpmodel(x, y, dm, optimizer)
    @show build_time
    
    solve_time = @elapsed JuMP.optimize!(m)
    println()
    @show solve_time
    
    return JuMP.objective_value(m)
end
```

Example output:
```
julia> @profile random_network_flow(150, Gurobi.Optimizer)
build_time = 0.288749231

Warning: excessive time spent in model updates.
Consider calling update less frequently.

Optimize a model with 150 rows, 22500 columns and 44700 nonzeros
Coefficient statistics:
  Matrix range     [1e+00, 1e+00]
  Objective range  [8e-03, 1e+00]
  RHS range        [4e-05, 1e-02]

Concurrent LP optimizer: primal simplex, dual simplex, and barrier
Showing barrier log only...

Presolve removed 1 rows and 150 columns
Presolve time: 0.02s
Presolved: 149 rows, 22350 columns, 44402 nonzeros

Ordering time: 0.00s

Barrier statistics:
 AA' NZ     : 1.103e+04
 Factor NZ  : 1.118e+04 (roughly 9 MBytes of memory)
 Factor Ops : 1.114e+06 (less than 1 second per iteration)
 Threads    : 4

Barrier performed 0 iterations in 0.04 seconds
Barrier solve interrupted - model solved by another algorithm


Solved with primal simplex
Solved in 1445 iterations and 0.04 seconds
Optimal objective  3.122653030e-02

solve_time = 2.96464323
0.031226530301749666

# for context, using GLPK
julia> random_network_flow(150, GLPK.Optimizer)
build_time = 0.112245802

solve_time = 0.216538412
0.031226530301749662
```

Notice that the overall solve time with Gurobi is order of magnitude worse than with GLPK, but of the ~3 seconds needed for `JuMP.optimize!` with Gurobi, only ~.04 seconds are spent in Gurobi! Gurobi also warns that excessive time is spent in model updates.

Basic profiling seems a little unreliable here, or I am bad at reading it - reported line numbers and functions don't always agree. That said, looks like lots of time is spent in `update_model!` of `grb_model.jl`.

odow:
Cleverly calling `update_model!` results in an order-of-magnitude improvement.

cc @evanfields

Closes #184.

`benchmark_jump.jl` is:
```julia
using Gurobi, JuMP
function benchmark(n = 150)
	m = JuMP.direct_model(Gurobi.Optimizer())
	@variable(m, f[1:n, 1:n] >= 0)
    return m
end
@time benchmark()
@time benchmark()
@time benchmark()
@time benchmark()
@time benchmark()
```
The results are
```
C:\Users\Oscar\.julia\dev\Gurobi>git checkout od/moi_speed
Switched to branch 'od/moi_speed'

C:\Users\Oscar\.julia\dev\Gurobi>julia --color=yes benchmark_jump.jl
  2.896283 seconds (5.20 M allocations: 279.623 MiB, 7.16% gc time)
  0.314600 seconds (821.42 k allocations: 63.965 MiB, 34.51% gc time)
  0.184337 seconds (802.34 k allocations: 62.967 MiB, 13.06% gc time)
  0.368996 seconds (802.34 k allocations: 62.967 MiB, 40.08% gc time)
  0.242605 seconds (802.34 k allocations: 62.967 MiB, 25.85% gc time)

C:\Users\Oscar\.julia\dev\Gurobi>git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.

C:\Users\Oscar\.julia\dev\Gurobi>julia --color=yes benchmark_jump.jl
  4.386719 seconds (5.49 M allocations: 294.247 MiB, 4.21% gc time)
  1.936523 seconds (802.34 k allocations: 62.966 MiB, 3.52% gc time)
  1.981247 seconds (821.42 k allocations: 63.966 MiB, 4.13% gc time)
  1.925684 seconds (802.34 k allocations: 62.967 MiB, 1.39% gc time)
  2.019717 seconds (802.34 k allocations: 62.967 MiB, 6.44% gc time)
```


evanfields:
I can confirm this branch gives a huge speedup on our research code - thanks! This is a really big deal for the problem we're working on.

Not sure I have much more to add. It seems like there is still meaningful overhead in solver communication; though surely some of that is inevitable. Is there a (distant) future where models can be passed at once rather than built incrementally, leading to further speedups?

odow:
> Gurobi would need to implement copy_to

LinQuadOptInterface.j should be the one to implement `copy_to`. The solver wrappers all implement plural arguments. I've opened an issue: https://github.com/JuliaOpt/LinQuadOptInterface.jl/issues/85.

src/MOIWrapper.jl
```diff
@@ -65,6 +76,28 @@ function Optimizer(;kwargs...)
...
+
+"""
+    _require_update(model::Optimizer)
+
+Calls `update_model!`, but only if the `model.needs_update` flag is set.
+"""
+function _update_if_necessary(model::Optimizer)
+    if model.needs_update
+        update_model!(model.inner)
+        model.needs_update = false
```
blegat: It will return false if `model.needs_update` and `nothing` otherwise if I am not mistaken. Maybe add an explicit return ?

PR #192, opened 2019-02-04
Update MPB_wrapper.jl

stefanobenati:
I added the line (line 760)

MPB.getsimplexiter(m::GurobiMathProgModel) = get_dblattr(m.inner, "IterCount")

to retrieve the number of simplex iteration. In my machine it works.


odow:
Great! Thanks @stefanobenati!

Typically we also add tests for any changes we make so that in future we don't break existing code, but since this is a small change and we can't automatically test Gurobi because of the commercial license, we'll make an exception. If you want to make more contributions we can show you how to change the code and add tests properly.

Thanks for contributing to JuliaOpt!

Issue #195, opened 2019-02-18
Index out of range for attribute `Obj`


ExpandingMan:
Hello all, I'm running on master as whatever performance issues exist for MOI are prohibitive for me on the tagged version.  I'm getting the following error on `optimize!`
```
ERROR: Gurobi.GurobiError(10006, "Index out of range for attribute 'Obj'")
Stacktrace:
 [1] set_dblattrarray!(::Gurobi.Model, ::String, ::Int64, ::Int64, ::Array{Float64,1}) at /home/savastio/.julia/dev/Gurobi/src/grb_attrs.jl:307
 [2] set_linear_objective!(::Gurobi.Optimizer, ::Array{Int64,1}, ::Array{Float64,1}) at /home/savastio/.julia/dev/Gurobi/src/MOI_wrapper.jl:366
 [3] unsafe_set!(::Gurobi.Optimizer, ::MathOptInterface.ObjectiveFunction{MathOptInterface.ScalarAffineFunction{Float64}}, ::MathOptInterface.ScalarAffineFunction{Float64}) at /home/savastio/.julia/packages/LinQuadOptInterface/ZMx9f/src/objective.jl:77
 [4] set at /home/savastio/.julia/packages/LinQuadOptInterface/ZMx9f/src/objective.jl:61 [inlined]
 [5] set(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Bridges.AllBridgedConstraints{Float64}}}, ::MathOptInterface.ObjectiveFunction{MathOptInterface.ScalarAffineFunction{Float64}}, ::MathOptInterface.ScalarAffineFunction{Float64}) at /home/savastio/.julia/packages/MathOptInterface/QVlW6/src/Bridges/bridgeoptimizer.jl:195
[...]

PR #196, opened 2019-02-18
Fix missing update

odow:
cc @ExpandingMan 

Can you check if this fixes #195 (and maybe provide a test-case?).

PR #198, opened 2019-02-20
Cache number of constraints to avoid unnecessary updates

odow:
@ExpandingMan can you check that this fixes https://github.com/JuliaOpt/Gurobi.jl/issues/195#issuecomment-465378059


ExpandingMan:
Nice!  This works, thanks so much!

Note that it's still looking about 10% to 20% slower (in direct mode) than in v0.18, so we might still have some regressions, but once we get this and #196 merged and the package tagged, I'm sure we'll be in good shape to fix the remaining regressions (I'll provide some benchmarks).  With any luck, perhaps the other PR's in LQOI will fix them.

odow:
> Note that it's still looking about 10% to 20% slower (in direct mode) than in v0.18

This is not surprising. This was one of the earliest wrappers written, so it has been through a few iterations of MOI. Now that things are settling down, I intend to fix a number of performance issues in LQOI (https://github.com/JuliaOpt/LinQuadOptInterface.jl/issues).

PR #210, opened 2019-05-24
Cache ObjVal and ObjBound attributes after solve

### LINKED ISSUE #209: [MOI] Changing bounds etc invalidates stored solution
I tried to use Gurobi with https://github.com/lanl-ansi/Juniper.jl, and ran into this bug
```julia
using Gurobi
const MOI = Gurobi.MOI

model = Gurobi.Optimizer()
x = MOI.add_variable(model)
c = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.GreaterThan(0.0))
MOI.set(model, MOI.ObjectiveFunction{MOI.SingleVariable}(), MOI.SingleVariable(x))
MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
MOI.optimize!(model)

# Can query the objective value
MOI.get(model, MOI.ObjectiveValue())

# Change a variable bound. The change is cached for update
MOI.set(model, MOI.ConstraintSet(), c, MOI.GreaterThan(1.0))

# Calling `get` forces an update
MOI.get(model, MOI.ConstraintSet(), c)

# Invalidating the stored objective :(
MOI.get(model, MOI.ObjectiveValue())
```
Error is
```Julia
ERROR: Gurobi.GurobiError(10005, "Unable to retrieve attribute 'ObjVal'")
```

I'm not sure what we should do, probably cache the objective value. And maybe objective bound, etc? The primal and dual solutions get cached by LQOI after `optimize!`, so these aren't affected.

cc @ccoffrin @Wikunia

odow:
Closes #209.

The bigger question at a MOI level is: should you still be able to query solution attributes after modifying the model?

cc @mlubin @blegat 

Tests passing locally
image

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -53,6 +53,11 @@ mutable struct Optimizer <: LQOI.LinQuadOptimizer
     num_variables::Int
     num_linear_constraints::Int
     num_quadratic_constraints::Int
+    # These fields cache the objective value and objective bound after the solve
+    # so that their values can still be queried, even if the model has been
+    # modified.
+    objective_value::Float64
+    objective_bound::Float64
 
     """
         Optimizer(env = nothing; kwargs...)
@@ -128,6 +133,9 @@ function MOI.empty!(model::Optimizer)
     model.num_variables = 0
     model.num_linear_constraints = 0
     model.num_quadratic_constraints = 0
+
+    model.objective_value = NaN
+    model.objective_bound = NaN
     return
 end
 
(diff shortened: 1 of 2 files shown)


blegat:
Good question, I am not sure what's best

mlubin:
I don't think we should guarantee this if it implies that a majority of solver wrappers have to add caching code for all solution attributes. After you modify a model, it's fair to not have access to attributes that are set by `optimize!`.

odow:
So then the follow-up question is: should we merge this or change Juniper? Other MIP solvers like GLPK work. I haven't tried CPLEX though.

blegat:
It seems better to change Juniper

PR #214, opened 2019-06-06
Update Julia to Julia 1.0

odow:
image

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,23 @@
+name = "Gurobi"
+uuid = "2e9cd046-0924-5485-92f1-d5272153d98b"
+
+[deps]
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+LinQuadOptInterface = "f8899e07-120b-5979-ab1d-7b97bb9e1a48"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MathProgBase = "fdba3010-5040-5b88-9595-932c9decdf73"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+
+[compat]
+LinQuadOptInterface = "~0.6.0"
+MathProgBase = "~0.5.0, ~0.6, ~0.7"
+julia = "≥ 1"
+
+[extras]
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[targets]
+test = ["Compat", "Pkg", "Random", "Test"]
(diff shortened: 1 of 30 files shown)


Project.toml
```diff
@@ -0,0 +1,23 @@
...
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+LinQuadOptInterface = "f8899e07-120b-5979-ab1d-7b97bb9e1a48"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MathProgBase = "fdba3010-5040-5b88-9595-932c9decdf73"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+
+[compat]
+LinQuadOptInterface = "~0.6.0"
+MathProgBase = "~0.5.0, ~0.6, ~0.7"
+julia = "≥ 1"
```
mlubin: "1" so that we don't claim to support Julia 2.0.

Project.toml
```diff
@@ -0,0 +1,23 @@
+name = "Gurobi"
+uuid = "2e9cd046-0924-5485-92f1-d5272153d98b"
```
mlubin: Missing repo and version.

src/MPB_wrapper.jl
```diff
@@ -116,7 +116,7 @@ function MPB.loadproblem!(m::GurobiMathProgModel, filename::AbstractString)
 end
 
 if VERSION >= v"0.7-"
```
mlubin: This version check can be removed.

src/grb_constrs.jl
```diff
@@ -79,7 +75,7 @@ end
 function add_constrs_t!(model::Model, At::Matrix{Float64}, rel::GCharOrVec, b::Vector)
     n, m = size(At)
     (m == length(b) && n == num_vars(model)) || error("Incompatible argument dimensions.")
-    add_constrs_t!(model, sparse(At), rel, b)
+    add_constrs_t!(model, SparseArrays.sparse(At), rel, b)
```
mlubin: `sparse` is used frequently enough that we may want `using SparseArrays`.

odow: I've left for now.

odow:
Once this is merged, I'm going to have a go at removing LQOI and writing a direct MOI v9 wrapper so see how much we can strip out.

Project.toml
```diff
@@ -11,7 +13,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 [compat]
 LinQuadOptInterface = "~0.6.0"
 MathProgBase = "~0.5.0, ~0.6, ~0.7"
-julia = "≥ 1"
+julia = "~1"
```
mlubin: I think "1" is sufficient to include all 1.x releases.

odow: I haven't quite got my head around this new format.

blegat:
> Once this is merged, I'm going to have a go at removing LQOI and writing a direct MOI v9 wrapper so see how much we can strip out.

What are the rationale for this ? The plan is to stop using LQOI from Gurobi or remove the need for LQOI everywhere ? I am not an expert of LQOI so I cannot judge what's the best solution. For SDOI, variable bridge will considerably simplify it so it may be possible to get rid of it after that.

odow:
The plan is to see how much more readable the Gurobi wrapper is without LQOI (one less layer). Now that MOI is more stable, a little duplication is probably better than an extra layer/dependency.

mlubin:
My impression is that many of the cases that motivated LQOI are now covered by bridges.

PR #216, opened 2019-06-10
Completely re-write the Gurobi MOI wrapper.

### LINKED ISSUE #144: MOIWrapper doesn't implement `get_quadratic_terms_objective` from LQOI
To reproduce, just set the model's objective to a ScalarQuadraticFunction and then try to get it: 

```julia
m = GurobiOptimizer()
MOI.set!(m, MOI.ObjectiveFunction{MOI.ScalarQuadraticFunction{Float64}}(),
        MOI.ScalarQuadraticFunction(MOI.ScalarAffineTerm{Float64}[], MOI.ScalarQuadraticTerm{Float64}[], -2.5))
MOI.get(m, MOI.ObjectiveFunction{MOI.ScalarQuadraticFunction{Float64}}())
```

gives: `MethodError: no method matching get_quadratic_terms_objective(::Gurobi.GurobiOptimizer)`

### LINKED ISSUE #201: Return MIP gap from gurobi, don't calculate it
https://github.com/JuliaOpt/Gurobi.jl/blob/ee3d837ea3c375b5dc2cca4dac61f6dd0a7541aa/src/MOI_wrapper.jl#L584-L588

Ref discussion in https://github.com/JuliaOpt/MathOptInterface.jl/issues/590

### LINKED ISSUE #203: Implement `MOI.Silent`
See http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/

### LINKED ISSUE #213: QCPDual values calculated but inaccessible using JuMP "dual" command
I have encountered what I think is a bug in Gurobi - JuMP link.

I am running Gurobi with the following call:
```julia
using Gurobi; m = Model(with_optimizer(Gurobi.Optimizer,QCPDual=1))
```
and the following lines return false and NaN:
```julia
println(has_duals(m))
println(dual(variable))
```
From logging in Julia terminal I can see Gurobi solves dual probelm but they are inaccessible in JuMP.
I can confirm Cplex and Xpress calculate them normally for this model and commands return true and value.
This problem only occurs with quadratic programming.

### LINKED ISSUE #215: Variables with a mixed primal start values
Problems with variables that have a mix of initial starting points and no initial starting points results in an exception error.  See

```julia
using JuMP
using Gurobi

m = Model()
@variable(m,x, lower_bound=0, upper_bound=2, start=0)
@variable(m,y, lower_bound=0, upper_bound=2)
JuMP.optimize!(m, with_optimizer(Gurobi.Optimizer))
```

This issue is likely related to this issue https://github.com/JuliaOpt/Ipopt.jl/issues/155

### LINKED ISSUE #217: Obtaining duals of convex problem with second order cone constraints not possible using Gurobi optimizer
I have a convex problem (which therefore should have duals) yet these don't seem to work. I can solve the problem just fine, but if I try to obtain the duals of a constraint I either get `NaN` or `AssertionError: dual >= 0.0`. Minimum working example below, I'm guessing it might be specific to Gurobi?

```julia
using JuMP
using Gurobi

m = Model(with_optimizer(Gurobi.Optimizer))

@variable(m, x >= 0)
@variable(m, y >= 0)
@variable(m, z >= 0)

@constraint(m, con1, x + y == 2)
@constraint(m, con2, x + y + z >= 0)
@constraint(m, con3, x^2 >= y^2 + z^2)

optimize!(m)

dual(con1)
dual(con2)
dual(con3)
```

odow:
All tests pass locally

image

Most recent benchmarks compared to current Gurobi master:
```
========== Results ==========

delete_constraint
BenchmarkTools.TrialJudgement: 
  time:   -96.03% => improvement (5.00% tolerance)
  memory: -98.63% => improvement (1.00% tolerance)

add_variable_constraints
BenchmarkTools.TrialJudgement: 
  time:   -32.44% => improvement (5.00% tolerance)
  memory: -36.64% => improvement (1.00% tolerance)

add_variables
BenchmarkTools.TrialJudgement: 
  time:   -20.63% => improvement (5.00% tolerance)
  memory: -31.15% => improvement (1.00% tolerance)

add_constraint
BenchmarkTools.TrialJudgement: 
  time:   -48.83% => improvement (5.00% tolerance)
  memory: -46.73% => improvement (1.00% tolerance)

add_variable
BenchmarkTools.TrialJudgement: 
  time:   -29.04% => improvement (5.00% tolerance)
  memory: -82.91% => improvement (1.00% tolerance)

delete_variable_constraint
BenchmarkTools.TrialJudgement: 
  time:   -52.91% => improvement (5.00% tolerance)
  memory: -76.74% => improvement (1.00% tolerance)

add_variable_constraint
BenchmarkTools.TrialJudgement: 
  time:   -52.85% => improvement (5.00% tolerance)
  memory: -54.30% => improvement (1.00% tolerance)

add_constraints
BenchmarkTools.TrialJudgement: 
  time:   -29.20% => improvement (5.00% tolerance)
  memory: -16.59% => improvement (1.00% tolerance)

delete_variable
BenchmarkTools.TrialJudgement: 
  time:   -35.54% => improvement (5.00% tolerance)
  memory: +6315.52% => regression (1.00% tolerance)
```
**!!! The massive increase in memory in `delete_variable` is due to the rebuild of the `OrderedDict` !!!**

Notable changes include:
 - removing the dependence on LQOI
 - more comprehensive coverage of MOI tests
 - removing VectorAffine support

<s>I found a couple of bugs in MOIT, will push a PR soon.</s>
- [x] Requires https://github.com/JuliaOpt/MathOptInterface.jl/pull/763
- [x] Requires https://github.com/JuliaOpt/MathOptInterface.jl/pull/767
- [x] Requires https://github.com/JuliaOpt/MathOptInterface.jl/pull/769

Todos
 - [x] RawParameter
 - [x] Indicator constraints
 - [x] BasisStatus
 - [x] benchmark compared to LQOI
 - [x] `add_constraints`?

Closes #217 
Closes #215 
Closes #213
Closes #203
Closes #201
Closes #144


test/MOI_wrapper.jl
```diff
@@ -3,107 +3,107 @@ const MOIT = MOI.Test
...
+        # @test_broken MOIT.start_values_test(Gurobi.Optimizer(GUROBI_ENV), OPTIMIZER)
+    end
+
+    @testset "supports_constrainttest" begin
+        MOIT.supports_constrainttest(OPTIMIZER, Float64, Int)
+    end
+
+    @testset "set_lower_bound_twice" begin
+        # Something weird is going on with the SplitInterval bridge.
+        # @test_broken MOIT.set_lower_bound_twice(OPTIMIZER, Float64)
```
odow: TODO: investigate the failure with bridges.

test/MOI_wrapper.jl
```diff
@@ -3,107 +3,107 @@ const MOIT = MOI.Test
...
+        MOIT.set_lower_bound_twice(Gurobi.Optimizer(GUROBI_ENV), Float64)
+    end
+
+    @testset "set_upper_bound_twice" begin
+        # Something weird is going on with the SplitInterval bridge.
+        # @test_broken MOIT.set_upper_bound_twice(OPTIMIZER, Float64)
```
odow: TODO: investigate the failure with bridges.

test/MOI_wrapper.jl
```diff
@@ -3,107 +3,107 @@ const MOIT = MOI.Test
...
+        )
+    end
+
+    @testset "scalar_function_constant_not_zero" begin
+        MOIT.scalar_function_constant_not_zero(OPTIMIZER)
+    end
+
+    @testset "start_values_test" begin
+        # We don't support ConstraintDualStart or ConstraintPrimalStart yet.
+        # @test_broken MOIT.start_values_test(Gurobi.Optimizer(GUROBI_ENV), OPTIMIZER)
```
odow: TODO: consider implementing ConstraintDualStart and ConstraintPrimalStart

chriscoey:
Are we saying bye bye to LQOI?

src/MOI_wrapper.jl
```diff
@@ -118,397 +210,1291 @@ end
 
 MOI.get(::Optimizer, ::MOI.SolverName) = "Gurobi"
 
-function MOI.empty!(model::Optimizer)
-    MOI.empty!(model, model.env)
-    setparam!(model.inner, "InfUnbdInfo", 1)
-    for (name, value) in model.params
-        setparam!(model.inner, name, value)
+function MOI.supports_constraint(::Optimizer, F::Type{<:MOI.AbstractFunction}, S::Type{<:MOI.AbstractSet})
+    return (F, S) in SUPPORTED_CONSTRAINTS
```
blegat: `MOI.supports_constraint(::Optimizer, ::Type{<:Union{...}}, ::Type{<:Union{...}}) = true` would be more efficient as Julia will be able to do constant propagation. The efficiency of `supports_constraint` is important for bridges

src/MOI_wrapper.jl
```diff
@@ -118,397 +210,1291 @@ end
...
+function MOI.get(model::Optimizer, ::MOI.ListOfConstraintAttributesSet)
+    return MOI.AbstractConstraintAttribute[MOI.ConstraintName()]
+end
+
+function indices_and_coefficients(model::Optimizer, f::MOI.ScalarAffineFunction{Float64})
+    f_canon = MOI.Utilities.canonical(f)
+    indices = Int[]
+    coefficients = Float64[]
+    for term in f_canon.terms
+        push!(indices, model[term.variable_index].column)
```
blegat: If no variable has been deleted, `model[term.variable_index].column` should be `term.variable_index.value` ? Needing a dictionary lookup for each term might be very slow
Related to https://github.com/JuliaOpt/LinQuadOptInterface.jl/issues/97

src/MOI_wrapper.jl
```diff
@@ -1,58 +1,110 @@
...
     needs_update::Bool
-    num_variables::Int
-    num_linear_constraints::Int
-    num_quadratic_constraints::Int
+
+    objective_type::ObjectiveType
+    is_feasibility::Bool
+
+    last_variable_index::Int
+    variable_info::Dict{MOI.VariableIndex, VariableInfo}
```
blegat: It would be more efficient to have a `Vector{Union{Nothing, AbstractBridge}}` if not too many variables are deleted.
See the `Map` in `src/Bridges/Variable/single_variable_optimizer.jl` in https://github.com/JuliaOpt/MathOptInterface.jl/pull/759.
We could maybe add a dictionary mapping from a bijection of `1:n` with holes to `T` in MOI:
```julia
struct VectorWithHoles{K, V} <: AbstractDict{K, V}
    f::Function # Map index -> key
    values::Vector{Union{Nothing, V}}
end
```
The elements of the dict are `[f(i) => values[i] for i in eachindex(values) if values[i] !== nothing]`.
Here you could use `VectorWithHoles{MOI.VariableIndex, VariableInfo}(MOI.VariableIndex, Union{Nothing, VariableInfo}[])`

src/MOI_wrapper.jl
odow: If `MOI.empty!` has been called, this this is also needed. Having the indices keep growing helped find that bug in MOI.Test (https://github.com/JuliaOpt/MathOptInterface.jl/pull/763).

odow:
> Are we saying bye bye to LQOI?

Not definitely, but I would like to. I think this wrapper is considerably more readable than previously (it has a lot less re-direction) at the cost of some duplication (2000 lines compared to 500 + LQOI).

Because of all the tests, it was pretty easy to implement, if a little tedious. It also enables Gurobi-specific work-arounds like the `needs_update` thing.

src/MOI_wrapper.jl
blegat: That's the purpose of xor'ing indices in MockOptimizer, we should have run the copy tests with it

src/MOI_wrapper.jl
odow: Re-above: this only makes sense if the `VariableIndex` always begin at `1` after `MOI.empty!`. I'll do some benchmarking to see how the current version compares to LQOI.

blegat: I do not expect this optimization to be slower than LQOI so this is out of the scope of this PR but it might help with
https://github.com/JuliaOpt/JuMP.jl/issues/1905
In direct mode, it might be a nonnegligible impact on performance

chriscoey:
Sounds reasonable. Hopefully it makes it easier to maintain and extend.

src/MOI_wrapper.jl
```diff
@@ -520,202 +1583,520 @@ function LQOI.get_primal_status(model::Optimizer)
...
+Internal function, not intended for external use.
+
+Change the linear constraint function at index `row` in `model` from
+`previous` to `replacement`. This function assumes that `previous` and
+`replacement` have exactly the same sparsity pattern w.r.t. which variables
+they include and that both constraint functions are in canonical form (as
+returned by `MOIU.canonical()`. Neither assumption is checked within the body
+of this function.
+"""
+function _replace_with_matching_sparsity!(
```
mlubin: Starting underscores are used inconsistently for internal functions.

src/MOI_wrapper.jl
```diff
@@ -118,397 +210,1368 @@ end
...
+
+function MOI.supports_constraint(
+    ::Optimizer, ::Type{MOI.VectorOfVariables}, ::Type{F}
+) where {F <: Union{MOI.SOS1{Float64}, MOI.SOS2{Float64}}}
+    return true
+end
+
+# We choose _not_ to support ScalarAffineFunction-in-Interval and
+# ScalarQuadraticFunction-in-Interval because Gurobi introduces some slack
+# variables that makes it hard to keep track of the column indices.
```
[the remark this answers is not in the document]
blegat: If it is internally bridged by Gurobi, that's the same as being bridged by the wrapper (for pure Julia solver, the wrapper and the interface are even the same). So the rule of : don't hardcode the bridge in the interface, just don't support it apply

odow:
Okay, now with an implementation of `CleverDict™`. Vector-backed storage so long as no elements are deleted. Keeps track of an offset so `MOI.empty!` works. Switches to `Dict` on `delete!`, and then back to vector on `empty!`.

src/MOI_wrapper.jl
```diff
@@ -1,58 +1,108 @@
...
+    # objects.
+
+    # Mappings from variable and constraint names to their indices. These are
+    # lazily built on-demand, so most of the time, they are `nothing`.
+    name_to_variable::Union{Nothing, Dict{String, MOI.VariableIndex}}
+    name_to_constraint_index::Union{Nothing, Dict{String, MOI.ConstraintIndex}}
+
+    # These two flags allow us to distinguish between FEASIBLE_POINT and
+    # INFEASIBILITY_CERTIFICATE when querying VariablePrimal and ConstraintDual.
+    has_unbounded_ray::Bool
```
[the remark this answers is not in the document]
odow: Yes. We either need to cache `has_unbounded_ray` and `has_infeasibility_ray` or `PrimalStatus` and `DualStatus` so when querying `VariablePrimal` or `ConstraintDual`, we don't need to do the check.

odow:
Now with a benchmarking script, that yields the following compared to `Gurobi#master` + `LQOI#master`.

`add_constraints` is slower because we haven't implemented that yet.

```
========== Results ==========

delete_constraint
BenchmarkTools.TrialJudgement: 
  time:   -53.10% => improvement (5.00% tolerance)
  memory: -76.74% => improvement (1.00% tolerance)

add_variable_constraint
BenchmarkTools.TrialJudgement: 
  time:   -47.47% => improvement (5.00% tolerance)
  memory: -85.41% => improvement (1.00% tolerance)

add_constraints
BenchmarkTools.TrialJudgement: 
  time:   +18.24% => regression (5.00% tolerance)
  memory: -55.91% => improvement (1.00% tolerance)
```

src/CleverDict.jl
```diff
@@ -0,0 +1,157 @@
...
+            if key.value == i + c.offset
+                continue
+            end
+            c.dict[index_to_key(K, i + c.offset)] = info
+        end
+        splice!(c.vector, key.value - c.offset)
+    else
+        for (i, v) in enumerate(c.vector)
+            if v == key
+                splice!(c.vector, i)
```
mlubin: This makes deletions `O(n)`. Is it feasible to set the `vector` to nothing if the dict is in use instead of keeping it updated?

odow: There is a trade-off:

 - making delete `O(n)`, and maintaining an ordered list of constraint indices
 - setting `vector` to `nothing`, and building an ordered list of variables on demand for calls like `get(model, ::ObjectiveFunction` or `get(model, ::ConstraintFunction)`, which needs a map between the Gurobi column and the MOI index.

Which would you prefer?

src/CleverDict.jl
```diff
@@ -0,0 +1,157 @@
...
+Provided no keys are deleted, the backing storage is a `Vector{V}`. Once a key
+has been deleted, the backing storage switches to a standard Julia `Dict{K,
+V}`.
+
+The i'th ordered element can be obtained with `c[LinearIndex(i)]`.
+
+Use `new_key(c::CleverDict)` to obtain the next key in the sequential order.
+
+When `empty!` is called, the largest index is stored in `offset` so that new
+keys are generated from `offset + 1`.
```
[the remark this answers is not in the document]
odow: To distinguish
```Julia
model = Gurobi.Optimizer()
x = MOI.add_variable(model)
MOI.empty!(model)
y = MOI.add_variable(model)
MOI.is_valid(model, x)
```
We don't have to do this though, but it caught a typo in MOI.Test

perf/perf.jl
```diff
@@ -0,0 +1,102 @@
...
+        MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(1.0, x), 0.0)
+    )
+    MOI.optimize!(model)
+    for con in cons
+        MOI.delete(model, con)
+    end
+    return model
+end
+suite["delete_constraint"] = @benchmarkable delete_constraint()
+
```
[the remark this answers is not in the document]
odow: I just took these from previous benchmarks. I'll add some more.

src/CleverDict.jl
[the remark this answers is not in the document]
mlubin: Doesn't seem worth it to me. There are other ways to fuzz the tests.

src/CleverDict.jl
mlubin: What about using `OrderedDict` instead of `Dict`?

src/CleverDict.jl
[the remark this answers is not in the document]
mlubin: Doesn't seem worth it to me. There are other ways to fuzz the tests.

src/CleverDict.jl
odow: So we could use an `OrderedDict`, but we would still need to build the list of variables on demand, because there isn't a mechanism to get the nth element of an OrderedDict (https://github.com/JuliaCollections/OrderedCollections.jl/blob/master/src/ordered_dict.jl).

odow:
Okay, things could do with a review again. I can't believe how much speed we left on the table with LQOI (see first post). The benchmarks definitely call for re-writing individual wrappers. It's also now much easier to test improvements via benchmarking. I wonder if we should roll these out via MOI.

src/CleverDict.jl
```diff
@@ -0,0 +1,177 @@
+import OrderedCollections
+
+"""
+    CleverDict{K, V}
```
mlubin: Consider adding a few unit tests for this data structure.

mlubin:
> It's also now much easier to test improvements via benchmarking. I wonder if we should roll these out via MOI.

Makes sense. The `CleverDict` could also go to MOIU if it's going to be used in multiple wrappers.

src/CleverDict.jl
blegat: It should probably be moved to MOI as it might be used by LQOI solvers (and maybe more).
In MOI, we currently use a `Vector{Union{Nothing, V}}` instead when we are in the use case of `CleverDict`. Not sure which one is better.

src/MOI_wrapper.jl
```diff
@@ -118,604 +207,2088 @@ end
...
+    model::Optimizer, c::MOI.ConstraintIndex{MOI.SingleVariable, <:Any}
+)
+    var_index = MOI.VariableIndex(c.value)
+    if haskey(model.variable_info, var_index)
+        return _info(model, var_index)
+    end
+    return throw(MOI.InvalidIndex(c))
+end
+
+function _throw_if_invalid(
```
blegat: You can use `MOI.throw_if_not_valid` instead

src/MOI_wrapper.jl
```diff
@@ -65,32 +109,77 @@ mutable struct Optimizer <: LQOI.LinQuadOptimizer
...
 
-function LQOI.LinearQuadraticModel(::Type{Optimizer}, env::Env)
-    # The user has passed an existing Env. Don't finalize it because we don't
-    # own this one.
-    return Model(env, "", finalize_env = false)
+function MOI.is_empty(model::Optimizer)
+    model.needs_update && return false
+    model.objective_type != SCALAR_AFFINE && return false
+    model.is_feasibility == false && return false
+    length(model.variable_info) != 0 && return false
```
blegat: isempty seems more appropriate

src/MOI_wrapper.jl
```diff
@@ -1,58 +1,102 @@
...
+    # Storage for constraint names associated with variables because Gurobi
+    # can only store names for variables and proper constraints.
+    # We can perform an optimization and only store three strings for the
+    # constraint names because, at most, there can be three SingleVariable
+    # constraints, e.g., LessThan, GreaterThan, and Integer.
+    lessthan_name::String
+    greaterthan_interval_or_equalto_name::String
+    type_constraint_name::String
+    function VariableInfo()
+        return new(MOI.VariableIndex(-1), 0, NONE, CONTINUOUS, "", "", "", "")
```
blegat: I would prefer VariableIndex(0) to signify an invalid one. Negative index means bridged. Even better would be to make the constructor with index and column arg

src/MOI_wrapper.jl
```diff
@@ -1,58 +1,99 @@
...
+    # variable type, and the names of SingleVariable-in-Set constraints.
+    variable_info::CleverDicts.CleverDict{MOI.VariableIndex, VariableInfo}
+
+    # An index that is incremented for each new constraint (regardless of type).
+    # We can check if a constraint is valid by checking if it is in the correct
+    # xxx_constraint_info. We should _not_ reset this to zero, since then new
+    # constraints cannot be distinguished from previously created ones.
+    last_constraint_index::Int
+    # ScalarAffineFunction{Float64}-in-Set storage.
+    affine_constraint_info::Dict{Int, ConstraintInfo}
```
joaquimg: Why these are not Clever ones?

joaquimg:
Thats an amazing work and benchmarks look very impressive!!!

How far from ready do you think it is?

How do you see implementing CPLEX, Xpress and GLPK interfaces from here? copy and pasting and figuring out mein differences?

What do you think about keeping all the interfaces performing well at (almost) the same time?
Maybe a Group for highlighting some improvements in a particular interface that could trigger issues for improvements in others?
Say someone sorts out some changes that would greatly improve performance, would the best way be opening issues in the related solvers?

I know this might end up being a greater discussion, just curious for thoughts :)

mtanneau:
`MOI.add_constraints` throws an error in the following case:
```julia
m = Gurobi.Optimizer()
n = 2

x = MOI.add_variables(m, n)
u = [1.0, 0.0]

MOI.add_constraints(m,
    [MOI.ScalarAffineFunction([MOI.ScalarAffineTerm(u[i], x[i])], 0.0) for i in 1:n],
    [MOI.EqualTo(0.0) for _ in 1:n]
)
```
outputs
```julia
ERROR: BoundsError: attempt to access 1-element Array{Int64,1} at index [2:2]
Stacktrace:
 [1] throw_boundserror(::Array{Int64,1}, ::Tuple{UnitRange{Int64}}) at ./abstractarray.jl:484
 [2] checkbounds at ./abstractarray.jl:449 [inlined]
 [3] view at ./subarray.jl:155 [inlined]
 [4] add_constraints(::Gurobi.Optimizer, ::Array{MathOptInterface.ScalarAffineFunction{Float64},1}, ::Array{MathOptInterface.EqualTo{Float64},1}) at /home/mtanneau/.julia/packages/Gurobi/86ICD/src/MOI_wrapper.jl:1248
 [5] top-level scope at none:0
```

It seems to happen when some of the input coefficients are zero.

mtanneau:
OK, so I think the error comes from the following:
* At line 1226, the call to `canonicalized_functions = MOI.Utilities.canonical.(f)` will drop the zero coefficients
https://github.com/JuliaOpt/Gurobi.jl/blob/od/moi9/src/MOI_wrapper.jl#L1226
* The number of non-zeros is computed based on `fi in canonicalized_functions` at line 1230, thus resulting in a lower number of non-zeros than there are coefficients in `f`
* Later, at line 1245, `fi` is actually enumerated based on the _original_ `f`, which means some coefficients may be explicit zeros, resulting in invalid indices as soon as one explicit zero is present.

Possible fixes:
* Count the explicit zeros when computing `nnz` in the `for` loop of l.1230
* Iterate through `canonicalized_functions` instead of `f` at l.1245:
```julia
for (i, (fi, si)) in enumerate(zip(canonicalized_functions, s))
```

src/MOI_wrapper.jl
```diff
@@ -118,604 +209,2102 @@ end
...
         return MOI.NO_SOLUTION
-    else
-        if stat == :optimal
-            return MOI.FEASIBLE_POINT
-        elseif stat == :solution_limit
-            return MOI.FEASIBLE_POINT
-        elseif (stat == :inf_or_unbd || stat == :infeasible) && has_dual_ray(model)
-            return MOI.INFEASIBILITY_CERTIFICATE
-        elseif stat == :suboptimal
-            r
```
mtanneau:
```julia
    if x ≈ ub
        return _dual_multiplier(model) * get_dblattrelement(model.inner, "RC", column)
    else
        return 0.0
    end
```
This strictly enforces complementary slackness, which may not be the case at the solver level (e.g. if you use an interior-point algorithm, or if you use larger numerical tolerances).
It's also not consistent with the case of `SingleVariable-in-Interval`.

IMO it would be preferable to return the +/- reduced cost, as is done for `SingleVariable-in-Interval`.

mtanneau: Why update if the variable is lower or upper bounded, but not if fixed or lower-and-upper bounded?

src/MOI_wrapper.jl
coroa: Because the `last_constraint_index` counter is shared over the types of constraints, so that there would be large holes in the CleverDict Array automatically.

Note, that this is not necessary, since the `i` in `ConstraintIndex{F,S}(i)` must only be unique per `F`, `S` type. It would probably be better to switch to `last_{affine,sos,quadratic}_constraint_index` and `CleverDict` for each of them.

coroa:
>     * Iterate through `canonicalized_functions` instead of `f` at l.1245:
> 
> ```julia
> for (i, (fi, si)) in enumerate(zip(canonicalized_functions, s))
> ```

Can confirm, that the latter fix works.

src/MOI_wrapper.jl
```diff
@@ -1,58 +1,99 @@
...
+    set::MOI.AbstractSet
+    # Storage for constraint names. Where possible, these are also stored in the
+    # Gurobi model.
+    name::String
+    ConstraintInfo(row::Int, set) = new(row, set, "")
+end
 
-mutable struct Optimizer <: LQOI.LinQuadOptimizer
-    LQOI.@LinQuadOptimizerBase(Model)
+mutable struct Optimizer <: MOI.ModelLike
```
coroa: Shouldn't `Optimizer` be a child of `MOI.AbstractOptimizer`, since it does expose `MOI.optimize!`?

src/MOI_wrapper.jl
odow: We probably don't need to.

odow: This was a deliberate choice in MOI. Maybe @mlubin or @blegat want to say more.

src/MOI_wrapper.jl
odow: Mainly because it's not performance critical. The only time we really need it is querying duals/ConstraintFunctions/etc. The extra hash doesn't matter. In contrast, we get column indices all the time, so it makes sense to make these fast.

odow:
image

blegat:
What is the `0`printed at the end ?

remi-garcia:
I'm reading this to learn and I found a typo, not exactly in the PR though
https://github.com/JuliaOpt/Gurobi.jl/blob/bf90918ab72943ffb06713d72db39a63ef65a362/src/MOI_wrapper.jl#L197-L208
`_require_update` -> `_update_if_necessary`

odow:
Thanks @remi-garcia. I've corrected.

joaquimg:
MOI 0.9 almost ready.
@odow do you have plans to take a last look in this PR before we can do the same for xpress and cplex?

odow:
I'll take a look today. You should be able take a start with Xpress. It should be sufficient to copy paste the whole thing, and 
then change out the Xpress specific calls.

odow:
Currently getting these errors:
```Julia
Name test with MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}: Test Failed at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:68
  Expression: MOI.get(model, MOI.VariableIndex, "Var1")
    Expected: Exception
  No exception thrown
Stacktrace:
 [1] macro expansion at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:68 [inlined]
 [2] macro expansion at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083 [inlined]
 [3] nametest(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}) at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:22
Name test with MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}: Test Failed at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:120
  Expression: MOI.get(model, MOI.ConstraintIndex, "Con2")
    Expected: Exception
  No exception thrown
Stacktrace:
 [1] nametest(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}) at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:120
 [2] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [3] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [4] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [5] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [6] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:48
Name test with MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}: Test Failed at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:121
  Expression: MOI.get(model, typeof(c2), "Con2")
    Expected: Exception
  No exception thrown
Stacktrace:
 [1] nametest(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}) at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:121
 [2] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [3] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [4] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [5] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [6] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:48
Name test with MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}: Test Failed at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:122
  Expression: MOI.get(model, typeof(cx), "Con2")
    Expected: Exception
  No exception thrown
Stacktrace:
 [1] nametest(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}) at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:122
 [2] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [3] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [4] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [5] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [6] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:48
Name test with MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}: Test Failed at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:154
  Expression: MOI.get(model, MOI.ConstraintIndex, "Con2")
    Expected: Exception
  No exception thrown
Stacktrace:
 [1] nametest(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}) at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:154
 [2] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [3] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [4] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [5] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [6] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:48
Name test with MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}: Test Failed at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:155
  Expression: MOI.get(model, typeof(c2), "Con2")
    Expected: Exception
  No exception thrown
Stacktrace:
 [1] nametest(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}) at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:155
 [2] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [3] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [4] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [5] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [6] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:48
Name test with MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}: Test Failed at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:156
  Expression: MOI.get(model, typeof(cx), "Con2")
    Expected: Exception
  No exception thrown
Stacktrace:
 [1] nametest(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}) at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:156
 [2] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [3] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [4] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
 [5] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [6] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:48
Name test with MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}: Error During Test at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:21
  Got exception outside of a @test
  MethodError: Cannot `convert` an object of type MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.EqualTo{Float64}} to an object of type String
  Closest candidates are:
    convert(::Type{T<:AbstractString}, !Matched::T<:AbstractString) where T<:AbstractString at strings/basic.jl:207
    convert(::Type{T<:AbstractString}, !Matched::AbstractString) where T<:AbstractString at strings/basic.jl:208
    convert(::Type{T}, !Matched::T) where T at essentials.jl:154
  Stacktrace:
   [1] setindex!(::Dict{String,MathOptInterface.ConstraintIndex}, ::String, ::MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.EqualTo{Float64}}) at ./dict.jl:373
   [2] set(::Gurobi.Optimizer, ::MathOptInterface.ConstraintName, ::MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.EqualTo{Float64}}, ::String) at /Users/oscar/.julia/dev/Gurobi/src/MOI_wrapper.jl:1348
   [3] set at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Bridges/bridge_optimizer.jl:619 [inlined]
   [4] macro expansion at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:171 [inlined]
   [5] macro expansion at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083 [inlined]
   [6] nametest(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}) at /Users/oscar/.julia/packages/MathOptInterface/mvlGI/src/Test/modellike.jl:22
   [7] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
   [8] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
   [9] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:59
   [10] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
   [11] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/MOI_wrapper.jl:48
   [12] include at ./boot.jl:326 [inlined]
   [13] include_relative(::Module, ::String) at ./loading.jl:1038
   [14] include(::Module, ::String) at ./sysimg.jl:29
   [15] include(::String) at ./client.jl:403
   [16] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/runtests.jl:17
   [17] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
   [18] top-level scope at /Users/oscar/.julia/dev/Gurobi/test/runtests.jl:17
   [19] include at ./boot.jl:326 [inlined]
   [20] include_relative(::Module, ::String) at ./loading.jl:1038
   [21] include(::Module, ::String) at ./sysimg.jl:29
   [22] include(::String) at ./client.jl:403
   [23] top-level scope at none:0
   [24] eval(::Module, ::Any) at ./boot.jl:328
   [25] exec_options(::Base.JLOptions) at ./client.jl:243
   [26] _start() at ./client.jl:436
Test Summary:                                                                       | Pass  Fail  Error  Broken  Total
MathOptInterface Tests                                                              | 2085     7      1       1   2094
  Unit Tests                                                                        |  807                         807
  Linear tests                                                                      |  517                         517
  Quadratic tests                                                                   |  190                         190
  Linear Conic tests                                                                |  135                         135
  Integer Linear tests                                                              |  139                         139
  ModelLike tests                                                                   |  213     7      1       1    222
    default_objective_test                                                          |                            No tests
    default_status_test                                                             |    3                           3
    nametest                                                                        |  103     7      1            111
      Name test with MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer} |  103     7      1            111
    validtest                                                                       |   11                          11
    emptytest                                                                       |   12                          12
    orderedindicestest                                                              |   13                          13
    copytest                                                                        |   37                          37
    scalar_function_constant_not_zero                                               |    2                           2
    start_values_test                                                               |    3                           3
    supports_constrainttest                                                         |    8                    1      9
    set_lower_bound_twice                                                           |   10                          10
    set_upper_bound_twice                                                           |   10                          10
  Gurobi Callback                                                                   |   10                          10
  LQOI Issue #38                                                                    |                            No tests
  User limit handling (issue #140)                                                  |    3                           3
  Constant objective (issue #111)                                                   |    4                           4
  Env                                                                               |   11                          11
  Conflict refiner                                                                  |   41                          41
  RawParameter                                                                      |    2                           2
  QCPDuals without needing to pass QCPDual=1                                        |   12                          12
  Add constraints                                                                   |    1                           1
ERROR: LoadError: Some tests did not pass: 2085 passed, 7 failed, 1 errored, 1 broken.
in expression starting at /Users/oscar/.julia/dev/Gurobi/test/runtests.jl:16
ERROR: Package Gurobi errored during testing
```

cc @blegat

odow:
image

odow:
image

PR #218, opened 2019-06-25
Implementing GRBgetvars

antonhinneck:
Hi,

I would like to include the function get_vars() (GRBgetvars).
It is similar to the already implemented function get_constrs(), but returns non-zero elements in a Gurobi.Model's constraint matrix based on variables.

---

While testing, I saw that get_constrs() produces inconsistent output regarding indices. The SparseArray it returns always is of size m (constraints) * n (variables). Say, a model has 4 constraints and 4 variables attached. 
If I want to get constraints 3-4, it will return a SparseArray (4 x 4) with those constraints at index [1, :] and [2,:] respectively.
If I use it to return all constraints (1-4), these previous constraints will be positioned at index [3, :] and [4,:].

The function get_constrs() always returns the constraints at their correct index if this little adjustment is made ...

**All tests pass.**

image

image

Best,

Anton

### DIFF
--- a/src/grb_constrs.jl
+++ b/src/grb_constrs.jl
@@ -208,7 +208,7 @@ function get_constrs(model::Model, start::Integer, len::Integer)
     V = Array{Float64}(undef, nnz)
     for i in 1:length(cbeg)-1
         for j in (cbeg[i]+1):cbeg[i+1]
-            I[j] = i
+            I[j] = i + start - 1
             J[j] = cind[j]+1
             V[j] = cval[j]
         end
--- a/src/grb_vars.jl
+++ b/src/grb_vars.jl
@@ -134,3 +134,86 @@ function del_vars!(model::Model, idx::Vector{Cint})
         throw(GurobiError(model.env, ret))
     end
 end
+
+function get_vars(model::Model, start::Integer, len::Integer)
+# http://www.gurobi.com/documentation/8.1/refman/c_grbgetvars.html
+
+    # NR_CONTRAINTS
+    #--------------
+    m = num_constrs(model)
+    # NR_VARIABLES
+    #-------------
+    n = num_vars(model)
+
+    # INPUT VALIDATION
+    #-----------------
+    @assert start > 0               "Indexing in Julia starts from 1."
+    @assert start <= n              string("Index out of bounds: There are only ", n, " variables attached to this model.")
+    @assert len > 0                 "At least one variable must be selected; len > 0."
+    @assert start * len <= m * n    "Maximal amount of possible non-zero elements surpassed."
+
+    # FUNCTION CALLS
+    #---------------
+    numnzP = Ref{Cint}()
+    vbeg = Array{Cint}(undef, len)
+
+    ret = @grb_ccall(getvars, Cint, (
+                        Ptr{Cvoid},     #Model
+                        Ptr{Cint},      #numnzP
+                        Ptr{Cint},      #vbeg
+                        Ptr{Cint},      #vind
+                        Ptr{Cdouble},   #vval
+                        Cint,           #start
+                        Cint),          #len
+                        model, numnzP, vbeg, C_NULL, C_NULL, Cint(start - 1), Cint(len))
+    if ret != 0
+        throw(GurobiError(model.env, ret))
+    end
+
+    nnz = numnzP[]
+    vind = Array{Cint}(undef, nnz)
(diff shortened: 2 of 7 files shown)


src/grb_vars.jl
```diff
@@ -134,3 +134,71 @@ function del_vars!(model::Model, idx::Vector{Cint})
...
+                        Cint),          #len
+                        model, numnzP, vbeg, vind, vval, Cint(start - 1), Cint(len))
+
+    if ret != 0
+        throw(GurobiError(model.env, ret))
+    end
+
+    for i in 1:size(vbeg, 1)
+        vbeg[i] += 1
+    end
```
odow: You should be able to go
```Julia
vbeg .+= 1
```

src/grb_vars.jl
```diff
@@ -134,3 +134,71 @@ function del_vars!(model::Model, idx::Vector{Cint})
...
+    V = Array{Float64}(undef, nnz)
+    for i in 1:length(vbeg) - 1
+        for j in vbeg[i]:vbeg[i + 1]
+            I[j] = vind[j]
+            J[j] = i + start - 1
+            V[j] = vval[j]
+        end
+    end
+
+    pop!(vbeg)
```
odow: No need for this.

test/C/test_constr_matrix.jl
```diff
@@ -0,0 +1,35 @@
+using Gurobi, Test
+
+@testset "constr_matrix" begin
+
+    simple_model_env = Gurobi.Env()
+    setparam!(simple_model_env, "OutputFlag", 0)
+
+    simple_model = Gurobi.Model(simple_model_env, "simple_mip", :maximize)
+
+    add_ivar!(simple_model, 0., 0, Inf)  # p1
```
odow: Please use `0.0` instead of `0.` or `0`

src/grb_constrs.jl
```diff
@@ -208,7 +208,7 @@ function get_constrs(model::Model, start::Integer, len::Integer)
     V = Array{Float64}(undef, nnz)
     for i in 1:length(cbeg)-1
         for j in (cbeg[i]+1):cbeg[i+1]
-            I[j] = i
+            I[j] = i + start - 1
```
odow: What is the implication of this? If anyone is using this, it will break existing code, correct? I'm tempted to allow it, because I don't think many (if any) people will depend on it, but maybe @mlubin has a different opinion.

antonhinneck: Yes, it would definitely mess up any dependencies.

test/C/test_read.jl
```diff
@@ -25,16 +25,16 @@ using Gurobi, Test
...
-    add_constr!(simple_model, [3., 5., 4., -1.], '<', 0.)
-    add_constr!(simple_model, [3., 5., 4., -1.], '>', 0.)
+    add_constr!(simple_model, [3., 5., 4., -1.], '<', 0.0)
+    add_constr!(simple_model, [3., 5., 4., -1.], '>', 0.0)
 
-    add_constr!(simple_model, [0.5, 2., 1., 0.], '<', 6.)
+    add_constr!(simple_model, [0.5, 2., 1., 0.0], '<', 6.)
 
     setparam!(simple_model, "Heuristics", 0.0)
     setparam!(simple_model, "Presolve", 0)
```
antonhinneck: I used the same model for testing that I used for another function - I also changed all "0." values to "0.0". Would it be better to make a single test file that contains a test sets for either function?

test/C/test_constr_matrix.jl
```diff
@@ -0,0 +1,36 @@
...
+
+    simple_model_env = Gurobi.Env()
+    setparam!(simple_model_env, "OutputFlag", 0)
+
+    simple_model = Gurobi.Model(simple_model_env, "simple_mip", :maximize)
+
+    add_ivar!(simple_model, 0.0, 0.0, Inf)  # p1
+    add_ivar!(simple_model, 0.0, 0.0, Inf)  # p2
+    add_ivar!(simple_model, 0.0, 0.0, Inf)  # p3
+    add_cvar!(simple_model, 1. , 0.0, Inf) # z
```
odow: Sorry, I wasn't clear enough. I mean add trailing zeros for all floating point constants, not just `0.`.

test/C/test_read.jl
odow: No, duplication in tests doesn't matter.

PR #231, opened 2019-08-14
Add VectorOfVariables-in-SecondOrderCone

odow:
There are a few todos:

In Gurobi.jl

- [x] Figure out how to set `t >= 0`

In MathOptInterface.jl

- [x] <s>Fix soc3 test to allow `INFEASIBLE_OR_UNBOUNDED` as a return https://github.com/JuliaOpt/MathOptInterface.jl/issues/836 https://github.com/JuliaOpt/MathOptInterface.jl/pull/839</s> We set `DualReductions=0`.
- [x] Implement `ConstraintFunction` and `ConstraintSet` in `geomean` bridge https://github.com/JuliaOpt/MathOptInterface.jl/issues/750

image

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -886,23 +886,30 @@ end
 """
     _set_variable_lower_bound(model, info, value)
 
-This function is used to indirectly set the lower bound of a variable. We need
-to do it this way, because if we add a SecondOrderCone constraint.
+This function is used to indirectly set the lower bound of a variable.
+
+We need to do it this way to account for potential lower bounds of 0.0 added by
+VectorOfVariables-in-SecondOrderCone constraints.
+
+See also `_get_variable_lower_bound`.
 """
 function _set_variable_lower_bound(model, info, value)
     if info.num_soc_constraints == 0
-        # No SOC constraints. Set directly.
+        # No SOC constraints, set directly.
         @assert isnan(info.lower_bound_if_soc)
         set_dblattrelement!(model.inner, "LB", info.column, value)
         _require_update(model)
     elseif value >= 0.0
-        # Regardless of whether there are SOC constraints, set directly.
+        # Regardless of whether there are SOC constraints, this is a valid bound
+        # for the SOC constraint and should over-ride any previous bounds.
         info.lower_bound_if_soc = NaN
         set_dblattrelement!(model.inner, "LB", info.column, value)
         _require_update(model)
     elseif isnan(info.lower_bound_if_soc)
-        # Previously, we had a +ve lower bound. We're setting this with a -ve
-        # one, but there are still some SOC constraints.
+        # Previously, we had a +ve lower bound (i.e., it was set in the case
+        # above). Now we're setting this with a -ve one, but there are still
+        # some SOC constraints, so we cache `value` and set the variable lower
+        # bound to `0.0`.
         @assert value < 0.0
         set_dblattrelement!(model.inner, "LB", info.column, 0.0)
         _require_update(model)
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -2525,3 +2554,140 @@ function MOI.supports(
...
+        error("Dimension of $(s) does not match number of terms in $(f)")
+    end
+    # SOC is the cone: t ≥ ||x||₂ ≥ 0. In quadratic form, this is
+    # t² - Σᵢ xᵢ² ≥ 0 and t ≥ 0.
+
+    # First, check the lower bound on t.
+
+    # TODO(odow): a better way to set the lower bound.
+    t_info = _info(model, f.variables[1])
+    set_dblattrelement!(model.inner, "LB", t_info.column, 0.0)
```
odow: Any ideas for how to do this? Because what if `t` already has a lower bound? I almost want to add a dummy variable with `t = t_dummy`, but then I'd have to account for the un-used variable elsewhere in the wrapper, leading to more complexity.

joaquimg: The MOI wrapper would know if there is an existing LB.
If there is one I would check what it is, if its less than zero, maybe, throw a warning?
If there is no LB we could add a the LB constraint OR only add this in the solver level and have an extra field in the optimizer to take car of this soc lb list of variables.

src/MOI_wrapper.jl
```diff
@@ -230,9 +235,26 @@ function MOI.supports_constraint(
     return true
 end
 
+function MOI.supports(
+    model, ::Optimizer, ::Union{MOI.ConstraintSet, MOI.ConstraintFunction},
```
blegat: That shouldn't be defined:
```julia
julia> using MathOptInterface

julia> const MOI = MathOptInterface
MathOptInterface

julia> MOI.is_copyable(MOI.ConstraintSet())
false
```

src/MOI_wrapper.jl
blegat:
```julia
if !MOI.is_valid(model, MOI.ConstraintIndex{MOI.SingleVariable, MOI.GreaterThan{Float64}(f.variables[1].value))
    # check that the bound is nonnegative, otherwise do
    MOI.add_constraint(model, MOI.SingleVariable(f.variables[1]), MOI.GreaterThan(0.0)) # This constraint should be hidden because it has not been added by the user. This is what `src/Bridges/bridge_optimizer.jl` does
end
```

src/MOI_wrapper.jl
odow: What's the rule for needing to `support` attributes? If `is_copyable` is `true`?

blegat: Yes :)
https://github.com/JuliaOpt/MathOptInterface.jl/blob/0ed068ff8342cf2501305963c22708b0b18a4b80/src/attributes.jl#L157-L159

odow:
I have a slightly ugly implementation that passes tests (of which I've added a lot to catch different edge cases. They should go into MOI). Opinions?

odow:
image

odow:
image

src/MOI_wrapper.jl
```diff
@@ -877,14 +883,70 @@ function MOI.delete(
...
+        set_dblattrelement!(model.inner, "LB", info.column, value)
+        _require_update(model)
+    elseif value >= 0.0
+        # Regardless of whether there are SOC constraints, this is a valid bound
+        # for the SOC constraint and should over-ride any previous bounds.
+        info.lower_bound_if_soc = NaN
+        set_dblattrelement!(model.inner, "LB", info.column, value)
+        _require_update(model)
+    elseif isnan(info.lower_bound_if_soc)
+        # Previously, we had a +ve lower bound (i.e., it was set in the case
```
mlubin: It's better to write out positive and negative. It took a few seconds for me to figure out what +ve and -ve meant.

PR #237, opened 2019-08-26
removed isascii assertions that were preventing unicode variables

### LINKED ISSUE #233: can we please get rid of the `isascii` assertion on variables?
I'm assuming the reason for this is that Gurobi only officially supports setting variable names to ASCII, but in Julia, where unicode is ubiquitous this is incredibly annoying, and clearly most things work perfectly fine even if you use non-ascii strings.

Could we simply delete the assertions in [`get_attrs.jl`](https://github.com/JuliaOpt/Gurobi.jl/blob/master/src/grb_attrs.jl)?

ExpandingMan:
I have fixed #233 and also JuMP issue [#2045](https://github.com/JuliaOpt/JuMP.jl/issues/2045).

I've added a simple test for unicode variable names, I can't say whether it will catch any potential issue.  I've also done a few simple tests of more complete problems with JuMP and haven't had any problems, so it looks good.  Of course, as Gurobi is proprietary and we don't have CI/CD, it wouldn't hurt if a maintainer pulled this and did their own checks.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -263,18 +263,11 @@ const SCALAR_SETS = Union{
 
 MOI.supports(::Optimizer, ::MOI.VariableName, ::Type{MOI.VariableIndex}) = true
 MOI.supports(::Optimizer, ::MOI.ConstraintName, ::Type{<:MOI.ConstraintIndex}) = true
-MOI.supports(::Optimizer, ::MOI.ObjectiveFunctionType) = true
 
 MOI.supports(::Optimizer, ::MOI.Name) = true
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -2046,7 +2046,7 @@ function MOI.get(
 end
 
 function MOI.get(model::Optimizer, ::MOI.ListOfConstraints)
-    constraints = Set{Any}()
+    constraints = Set{Tuple{DataType,DataType}}()
```
blegat: This is already done in https://github.com/JuliaOpt/Gurobi.jl/pull/235

PR #251, opened 2019-09-14
Fix handling of duplicate names

### LINKED ISSUE #249: Duplicate variable names
Giving the same name to two different variables does not raise an error (as per [MOI docs](http://www.juliaopt.org/MathOptInterface.jl/stable/apireference/#MathOptInterface.VariableName)).
However, doing so leads to an error being raised when looking up _any_ variable by name afterwards.

Is that a feature or a bug?

```julia
import Gurobi  # Gurobi master
import MathOptInterface  # MOI 0.9.2
MOI = MathOptInterface

m = Gurobi.Optimizer()
x = MOI.add_variable(m)
y = MOI.add_variable(m)
z = MOI.add_variable(m)

MOI.set(m, MOI.VariableName(), x, "x")
MOI.set(m, MOI.VariableName(), y, "x")  # duplicate name, does not raise an error
MOI.set(m, MOI.VariableName(), z, "z")

MOI.get(m, MOI.VariableIndex, "x")  # raises an error due to duplicate names
MOI.get(m, MOI.VariableIndex, "z")  # raises the same error. Should it?
```
Error message:
```julia
ERROR: Duplicate variable name detected: x
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] _rebuild_name_to_variable(::Gurobi.Optimizer) at /home/mathieu/GitHub/Gurobi.jl/src/MOI_wrapper.jl:471
 [3] get(::Gurobi.Optimizer, ::Type{MathOptInterface.VariableIndex}, ::String) at /home/mathieu/GitHub/Gurobi.jl/src/MOI_wrapper.jl:458
 [4] top-level scope at none:0
```

odow:
Simplified (and corrected) handling of variable and constraint names. Note that instead of trying to maintain the name-to-X mapping, we just throw it away if anything changes. This is inefficient if your workflow is add-query-add-query-etc. But it should be fine for the most common work-flow of add a lot of things - query.

image

Fixes #249

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -93,8 +93,8 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
 
     # Mappings from variable and constraint names to their indices. These are
     # lazily built on-demand, so most of the time, they are `nothing`.
-    name_to_variable::Union{Nothing, Dict{String, MOI.VariableIndex}}
-    name_to_constraint_index::Union{Nothing, Dict{String, MOI.ConstraintIndex}}
+    name_to_variable::Union{Nothing, Dict{String, Union{Nothing, MOI.VariableIndex}}}
+    name_to_constraint_index::Union{Nothing, Dict{String, Union{Nothing, MOI.ConstraintIndex}}}
 
     # These two flags allow us to distinguish between FEASIBLE_POINT and
     # INFEASIBILITY_CERTIFICATE when querying VariablePrimal and ConstraintDual.
@@ -456,6 +456,9 @@ function MOI.delete(model::Optimizer, v::MOI.VariableIndex)
         end
     end
     model.name_to_variable = nothing
+    # We throw away name_to_constraint_index so we will rebuild SingleVariable
+    # constraint names without v.
+    model.name_to_constraint_index = nothing
     return
 end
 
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -450,27 +450,37 @@ function MOI.delete(model::Optimizer, v::MOI.VariableIndex)
...
-    model.name_to_variable = Dict{String, MOI.VariableIndex}()
+    model.name_to_variable = Dict{String, Union{Nothing, MOI.VariableIndex}}()
     for (index, info) in model.variable_info
         if info.name == ""
             continue
         end
         if haskey(model.name_to_variable, info.name)
-            model.name_to_variable = nothing
-            error("Duplicate variable name detected: $(info.name)")
+            model.name_to_variable[info.name] = nothing
```
blegat: We use `MOI.VariableIndex(0)` in `MOIU.Model`

src/MOI_wrapper.jl
```diff
@@ -1409,13 +1401,14 @@ end
...
 function _rebuild_name_to_constraint_index_util(model::Optimizer, dict, F)
     for (index, info) in dict
-        info.name == "" && continue
-        if haskey(model.name_to_constraint_index, info.name)
-            model.name_to_constraint_index = nothing
-            error("Duplicate constraint name detected: $(info.name)")
+        if info.name == ""
+            continue
+        elseif haskey(model.name_to_constraint_index, info.name)
+            model.name_to_constraint_index[info.name] = nothing
```
blegat: We use `MOI.ConstraintIndex{Nothing, Nothing}(0)` for `MOIU.Model`

mlubin:
This makes me think we should remove lookup by name from MOI and move it to JuMP. It doesn't make sense for all the solver wrappers to have to reimplement this logic.

odow:
Except that it is used heavily in the MOI tests when we parse models from strings. And MOI format.

blegat:
I feel that we might avoid duplicate code with utilities such as https://github.com/JuliaOpt/MathOptInterface.jl/blob/98a0e472e1757db14b849bbba8fe84597856def2/src/Utilities/model.jl#L263-L282
We can relax the behavior of name lookup in DIRECT mode (e.g. Mosek throws an error when setting a duplicate name, not when querying it).

src/MOI_wrapper.jl
odow: I'd prefer to keep `nothing`. It's more explicit, and the small union shouldn't be a performance bottleneck.

PR #252, opened 2019-09-14
Use tolerance for SingleVariable dual checks

### LINKED ISSUE #250: Remove `x ≈ bnd` checks when querying `ConstraintDual` for `SingleVariable` constraints
When querying the dual variable associated to a `MOI.SingleVariable`-in-`S`, the wrapper first check whether the corresponding variable is at its lower/upper bound, e.g.:
```julia
    if x ≈ ub
        return _dual_multiplier(model) * get_dblattrelement(model.inner, "RC", column)
    else
        return 0.0
    end
```
(the offending code is for [`LessThan`](https://github.com/JuliaOpt/Gurobi.jl/blob/master/src/MOI_wrapper.jl#L1846) and [`GreaterThan`](https://github.com/JuliaOpt/Gurobi.jl/blob/master/src/MOI_wrapper.jl#L1860))

I would argue it is dangerous to do so, and may lead to returning incorrect values. Specifically:
* ~Even if a variable is at its lower/upper bound, it may still be basic (#degeneracy), and thus may have a non-zero reduced cost.~ [edit: I was wrong, this is indeed not a problem]
* More problematic, if the lower/upper bound is zero, then `x ≈ ub` will evaluate to `false` whenever `x` is not _exactly_ zero. This happens all the time when solving with barrier and no crossover, and more generally because of numerical tolerances.
* For `x-in-Interval`, there is no bound check ([here](https://github.com/JuliaOpt/Gurobi.jl/blob/master/src/MOI_wrapper.jl#L1881))

odow:
Any ideas for how to test this in a numerically stable way?

Closes #250

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -1882,8 +1882,7 @@ function MOI.get(
     model::Optimizer, ::MOI.ConstraintDual,
     c::MOI.ConstraintIndex{MOI.SingleVariable, MOI.LessThan{Float64}}
 )
-    column = _info(model, c).column
-    reduced_cost = get_dblattrelement(model.inner, "RC", column)
+    reduced_cost = get_dblattrelement(model.inner, "RC", _info(model, c).column)
     sense = MOI.get(model, MOI.ObjectiveSense())
     if sense == MOI.MIN_SENSE && reduced_cost < 0
         return reduced_cost
@@ -1899,8 +1898,7 @@ function MOI.get(
     model::Optimizer, ::MOI.ConstraintDual,
     c::MOI.ConstraintIndex{MOI.SingleVariable, MOI.GreaterThan{Float64}}
 )
-    column = _info(model, c).column
-    reduced_cost = get_dblattrelement(model.inner, "RC", column)
+    reduced_cost = get_dblattrelement(model.inner, "RC", _info(model, c).column)
     sense = MOI.get(model, MOI.ObjectiveSense())
     if sense == MOI.MIN_SENSE && reduced_cost > 0
         return reduced_cost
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -1850,7 +1850,7 @@ function MOI.get(
     column = _info(model, c).column
     x = get_dblattrelement(model.inner, "X", column)
     ub = get_dblattrelement(model.inner, "UB", column)
-    if x ≈ ub
+    if isapprox(x, ub, atol = get_dbl_param(model.inner, "FeasibilityTol"))
```
blegat: What does `get_dblattrelement(model.inner, "RC", column)` return if `x` is different from `ub` ?

odow: See #250. It depends on the lower bound.

odow:
image

src/MOI_wrapper.jl
```diff
@@ -1882,12 +1882,14 @@ function MOI.get(
...
     c::MOI.ConstraintIndex{MOI.SingleVariable, MOI.LessThan{Float64}}
 )
-    column = _info(model, c).column
-    x = get_dblattrelement(model.inner, "X", column)
-    ub = get_dblattrelement(model.inner, "UB", column)
-    if x ≈ ub
-        return _dual_multiplier(model) * get_dblattrelement(model.inner, "RC", column)
+    reduced_cost = get_dblattrelement(model.inner, "RC", _info(model, c).column)
+    sense = MOI.get(model, MOI.ObjectiveSense())
+    if sense == MOI.MIN_SENSE && reduced_cost < 0
```
mlubin: This could use an explanatory comment also.

mtanneau:
> Any ideas for how to test this in a numerically stable way?

How about solving a simple model, say
```math
min x
s.t. l <= x <= u
```
once written with variable bounds (`MOI.SingleVariable`), the other written with general linear constraints (`MOI.ScalarAffineFunction`). The duals of both models should be identical (up to numerical tolerances).

odow:
See: https://github.com/JuliaOpt/MathOptInterface.jl/pull/883

PR #257, opened 2019-09-23
Handle GRB_UNDEFINED start values

### LINKED ISSUE #241: Setting `GRB_UNDEFINED` start values
The following branch is missing an `else`. If we pass `nothing`, we should probably set `GRB_UNDEFINED`. See https://www.gurobi.com/documentation/8.1/refman/start.html#attr:Start

https://github.com/JuliaOpt/Gurobi.jl/blob/db91e1e9fabc3f62f30265780e6a577016ebba4e/src/MOI_wrapper.jl#L1957-L1960

odow:
Closes #241

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -2009,17 +2009,17 @@ function MOI.set(
 )
     info = _info(model, x)
     info.start = value
-    if value !== nothing
-        set_dblattrelement!(model.inner, "Start", info.column, value)
-        _require_update(model)
-    end
+    grb_value = value !== nothing ? value : GRB_UNDEFINED
+    set_dblattrelement!(model.inner, "Start", info.column, grb_value)
+    _require_update(model)
     return
 end
 
 function MOI.get(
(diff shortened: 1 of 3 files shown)


blegat:
We should have a test, if Gurobi change the value of the constant we won't notice it otherwise

odow:
image

src/MOI_wrapper.jl
```diff
@@ -2009,17 +2009,17 @@ function MOI.set(
...
+    _require_update(model)
     return
 end
 
 function MOI.get(
     model::Optimizer, ::MOI.VariablePrimalStart, x::MOI.VariableIndex
 )
-    return _info(model, x).start
+    grb_value = _info(model, x).start
+    return grb_value === GRB_UNDEFINED ? nothing : grb_value
```
blegat: The `start` field of `VariableInfo` is either `nothing` or the value of the starting value, it is never set to `GRB_UNDEFINED` if I'm not mistaken.

PR #258, opened 2019-09-24
DNMY: Implement MOI callbacks

odow:
Implements the MOI callbacks added https://github.com/JuliaOpt/MathOptInterface.jl/pull/782

Requires:
 - [x] a new tag of MOI and a minimum bound

TODOs
 - [x] Tests for `OptimizeInProgress`
 - [x] Set a flag for the different callbacks to check if, for example, `UserCut` is added in `LazyConstraint`
 - [x] Testing. Not obvious how to do well.

image


odow:
I would prefer that we keep as-is for now. Once we implement other solvers, if they turn out differently we can iterate.

image

odow:
The tests now have nice examples using `Gurobi.CallbackFunction`, as well as `LazyConstraintCallback`, `UserCutCallback`, and `HeuristicCallback` to add lazy constraints, user cuts, and heuristic solutions respectively. It demonstrates how you can use Gurobi-specific function in the `Gurobi.CallbackFunction` case.

test/MOI_callbacks.jl
```diff
@@ -0,0 +1,397 @@
...
+            MOI.submit(
+                model,
+                MOI.UserCut(cb_data),
+                MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(1.0, x), 0.0),
+                MOI.LessThan(5.0)
+            )
+        end)
+        @test_throws(
+            ErrorException(
+                "`MOI.UserCut` can only be called from UserCutCallback."
```
blegat: We should have an MOI error for that. It's likely to be thrown from every solver with a solver-independent error.

blegat: We could even have a default fallback
```julia
# in MOI
MOI.submit(::MOI.ModelLike, ::UserCut{<:AbstractLazyConstraintCallbackData}, args...) = throw(...)
# in Gurobi
struct UserCutCallbackData ... end
MOI.submit(::Optimizer, ::UserCut{UserCutCallbackData}, ...) = ...
```

odow: :+1: on a MOI error.

:-1: on different callback data for each callback data, because that doesn't solve
```Julia
MOI.set(model, MOI.LazyConstraintCallback(), cb_data -> begin
    MOI.submit(model, MOI.UserCut{Gurobi.UserCutCallbackData}(cb_data), f, s)
end)
```

blegat: The idea is that the type of `cb_data` will be `LazyConstraintCallbackData` so the user will have to do:
```julia
MOI.set(model, MOI.LazyConstraintCallback(), cb_data -> begin
    MOI.submit(model, MOI.UserCut(Gurobi.UserCutCallbackData(cb_data.inner_data))), f, s)
end)
```
With the current proposal, he can also do
```julia
MOI.set(model, MOI.LazyConstraintCallback(), cb_data -> begin
    model.callback_state = Gurobi.CB_USER_CUT
    MOI.submit(model, MOI.UserCut(cb_data), f, s)
end)
```

odow: > With the current proposal, he can also do

Sure. But I can do anything modifying internal fields. We still need the `model.callback_state` field to catch `OptimizeInProgress`, so the choice is between making this a `Bool` and having lots of different `XXXCallbackData` types, or making it an enum.

I'll stick with the enum for now.

blegat: Let's keep it simple for now on this PR I agree. This thread is more at the stage of brainstorming future ideas as it won't have impact on user facing interface.

odow:
Here's an example using callbacks with JuMP:

```Julia
using JuMP, Gurobi, Test

macro lazy_constraint(model, cb_data, constr)
    model = esc(model)
    cb_data = esc(cb_data)
    quote
        con = $(esc(Expr(
            :macrocall,
            Symbol("@build_constraint"),
            LineNumberNode(@__LINE__, @__FILE__),
            constr
        )))
        MOI.submit(
            backend($(model)),
            MOI.LazyConstraint($(cb_data)),
            JuMP.moi_function(con.func),
            con.set
        )
    end
end

function lazy_constraint_test()
    model = direct_model(
        Gurobi.Optimizer(
            OutputFlag = 0,
            Cuts = 0,
            Presolve = 0,
            Heuristics = 0,
            LazyConstraints = 1
        )
    )
    @variable(model, 0 <= x <= 2.5, Int)
    @variable(model, 0 <= y <= 2.5, Int)
    @objective(model, Max, y)
    MOI.set(model, MOI.LazyConstraintCallback(), (cb_data) -> begin
        x_val = MOI.get(model, MOI.CallbackVariablePrimal(cb_data), x)
        y_val = MOI.get(model, MOI.CallbackVariablePrimal(cb_data), y)
        if y_val - x_val > 1 + 1e-6
            @lazy_constraint(model, cb_data, y - x <= 1)
        elseif y_val + x_val > 3 + 1e-6
            @lazy_constraint(model, cb_data, x + y <= 3)
        end
    end)
    optimize!(model)
    @test value(x) == 1
    @test value(y) == 2
end

lazy_constraint_test()
```

odow:
image

odow:
cc @chengg04 :)

PR #265, opened 2019-10-24
Gurobi 9.0 beta compatiblity.

bhalonen:


odow:
Is there a list of breaking changes?

bhalonen:
Yeah, I think I am not supposed to disclose these things until they officially release. Just read the agreement. I'll close this.

odow:
I'm aware that they are making some large changes. I would suggest that you make the necessary changes locally without pushing to GitHub. Send me an email privately if you run into any issues.

PR #267, opened 2019-11-05
Add upper bound on OrderedCollections

blegat:


odow:
Where do we even use OrderedCollections? `git blame` suggests I added it when I re-wrote the MOI wrapper, so it was probably before we moved the implementation into `CleverDict`. 

I wish Julia was like Go and threw an error if `Project.toml` includes an un`include`d package.

PR #262, opened 2019-10-14
Fix int double error when quering NodesCount

LefterisM:
Edit(odow): Closes #261

PR #275, opened 2019-12-03
Fix types of attribute queries and add tests

odow:
Closes #262

@LefterisM, I cherry-picked your commit from #262.


odow:
image

PR #282, opened 2020-01-03
Specialization of MOI.delete for vector of variables 

henriquebecker91:
Arised from a discussion with @odow and @mlubin.

MOI.delete already has a fallback method for a vector of variable parameters (which just calls the single variable delete over each variable of the vector). The problem is that this fallback has O(n^2) performance the way the things are now (i.e., needing to update the model after each single variable deletion to keep the correspondence between MOI indexes and internal gurobi indexes). This specialization fixes the problem in an easy way but limited way (a better but much more complicated solution would be restructuring the whole lazy update implementation to keep a map between MOI indexes and internal Gurobi indexes).

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -460,6 +460,27 @@ function MOI.is_valid(model::Optimizer, v::MOI.VariableIndex)
     return haskey(model.variable_info, v)
 end
 
+function MOI.delete(model::Optimizer, indices::Vector{<:MOI.VariableIndex})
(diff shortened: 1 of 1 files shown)


odow:
Great! Can you post a screenshot confirming that the tests you wrote for MOI pass with this branch?

henriquebecker91:
Do you mean like this?

[](https://postimg.cc/KRHH0mnw)

I was not sure of what you wanted, so if this is not what you want please correct me.

PR #300, opened 2020-03-27
Draft of batch constraint deletion. Preexisting tests passing.

henriquebecker91:
Discussed in Gitter. This should help to avoid the "Warning: excessive time spent in model updates. Consider calling update less frequently." message when multiple constraints are deleted in a single batch. This is a draft and is not yet fully tested, and probably changes will be necessary.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -916,14 +916,13 @@ function MOI.add_constraints(
     return indices
 end
 
-function MOI.delete(
+function _delete_no_update(
     model::Optimizer,
     c::MOI.ConstraintIndex{MOI.SingleVariable, MOI.LessThan{Float64}}
 )
     MOI.throw_if_not_valid(model, c)
     info = _info(model, c)
     set_dblattrelement!(model.inner, "UB", info.column, Inf)
-    _require_update(model)
     if info.bound == LESS_AND_GREATER_THAN
         info.bound = GREATER_THAN
     else
@@ -932,6 +931,51 @@ function MOI.delete(
     info.lessthan_name = ""
     model.name_to_constraint_index = nothing
     return
+end
+
+function MOI.delete(
+    model::Optimizer,
+    c::MOI.ConstraintIndex{MOI.SingleVariable, MOI.LessThan{Float64}}
+)
+    _delete_no_update(model, c)
+    _require_update(model)
+    return
+end
+
+# batch version
+function MOI.delete(
+    model::Optimizer,
+    cs::Vector{MOI.ConstraintIndex{MOI.SingleVariable, MOI.LessThan{Float64}}}
+)
+    _delete_no_update.(model, cs)
+    _require_update(model)
+    return
+end
+
+function _set_variable_lower_bound_no_update(model, info, value)
+    if info.num_soc_constraints == 0
+        # No SOC constraints, set directly.
+        @assert isnan(info.lower_bound_if_soc)
+        set_dblattrelement!(model.inner, "LB", info.column, value)
+    elseif value >= 0.0
+        # Regardless of whether there are SOC constraints, this is a valid bound
+        # for the SOC constraint and should over-ride any previous bounds.
+        info.lower_bound_if_soc = NaN
+        set_dblattrelement!(model.inner, "LB", info.column, value)
+    elseif isnan(info.lower_bound_if_soc)
+        # Previously, we had a non-negative lower bound (i.e., it was set in the
+        # case above). Now we're setting this with a negative one, but there are
+        # still some SOC constraints, so we cache `value` and set the variable
+        # lower bound to `0.0`.
+        @assert value < 0.0
+        set_dblattrelement!(model.inner, "LB", info.column, 0.0)
+        info.lower_bound_if_soc = value
+    else
+        # Previously, we had a negative lower bound. We're setting this with
+        # another negative one, but there are still some SOC constraints.
+        @assert info.lower_bound_if_soc < 0.0
+        info.lower_bound_if_soc = value
+    end
 end
 
 """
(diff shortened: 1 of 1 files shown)


odow:
So it isn't obvious where we're updating the model in the existing code? (i.e., I expected to see an `update_if_required(model)` in the code you deleted.)

Do you have a simple benchmark that demonstrates the performance issue?

henriquebecker91:
This first commit was a brain fart, I confused `_require_update` and `_update_if_required` and as all those methods are used in the part my code is slowed, I thought it was very clear what the problem was, and I was wrong. I will revert those changed in next commit. It was a good idea to mark this as a draft.

However, my code is in fact slowed in the same part these methods is used, I have added a lot of `@timeit`s from [TimerOutputs](https://github.com/KristofferC/TimerOutputs.jl) to help me found the problem, and this is a excerpt of the output:

```
──────────────────────────────────────────────────────────────────────────────
                                       Time                   Allocations      
                               ──────────────────────   ───────────────────────
       Tot / % measured:        18446736165s / 0.00%        10.1GiB / 64.7%    

 Section               ncalls     time   %tot     avg     alloc   %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 run                        1    97.1s   100%   97.1s   6.56GiB  100%   6.56GiB
   build_model              1    84.8s  87.4%   84.8s   3.33GiB  50.7%  3.33GiB
     restricted_final       1    83.9s  86.4%   83.9s   2.93GiB  44.7%  2.93GiB
       relax_cm             1    48.9s  50.4%   48.9s   21.2MiB  0.32%  21.2MiB
       discrete_opt...      1    19.8s  20.4%   19.8s     0.00B  0.00%    0.00B
       relaxed_opti...      1    11.4s  11.8%   11.4s     0.00B  0.00%    0.00B
```

The last two lines are solving two distinct subsets of the model (and at one moment the model is entirely relaxed and at other some variables are integer/binary), however it is the line `relax_cm` that takes 50% of the total time. This line is a broadcast of the method below over a large number of variables:
```
function save_and_relax!(var :: VariableRef)
  svc = SavedVarConf(var)
  if svc.was_int
    unset_integer(var)
  elseif svc.was_bin
    unset_binary(var)
    if !svc.had_lb || svc.lb < 0.0 
      set_lower_bound(var, 0.0)
    end 
    if !svc.had_ub || svc.ub > 1.0 
      set_upper_bound(var, 1.0)
    end 
  end 

  return svc 
end
```
The `SavedVarConf` just queries info, so the culprit is probably `unset_integer` but I will check this with more detail after lunch, just keeping you updated. The idea would be that this method received a vector of variables and called vector versions of the methods used internally that are slow (if broadcast works for some of them, those could be broadcasts, but with this time, at least one of them, probably `unset_integer`, needs a batch version, if possible).

odow:
So to understand what's going on, we really need a minimum reproducible example.

If `SavedVarConf` queries variable information, then this is probably the culprit. You could query everything first, then loop through changing bounds. 

For example:
```julia
svcs = SavedVarConf.(all_variables(model))
for (x, svc) in zip(all_variables(model), svcs)
    save_and_relax!(x, svc)
end
```

Query, change, query, change is the issue.

We're never going to be able to avoid this with Gurobi.jl. It has particular update semantics that mean there are always going to be work-flows that hit the slow path.

henriquebecker91:
```julia
using JuMP
import Gurobi

function alternating()
    m = direct_model(Gurobi.Optimizer())
    @variable(m, x[1:100000] >= 0, Int)
    for e in x
        unset_integer(e)
        #is_integer(e)
        #has_upper_bound(e)
        lower_bound(e)
    end
    return
end

function batch()
    m = direct_model(Gurobi.Optimizer())
    @variable(m, x[1:100000] >= 0, Int)
    unset_integer.(x)
    is_integer.(x)
    #is_integer.(x)
    #has_upper_bound.(x)
    lower_bound.(x)
    return
end

function test()
    println("-------------------- batch --------------------------")
    @time batch()
    println("-------------------- alternating --------------------")
    @time alternating()
end

test()
println("\n-------------------- second run ---------------------\n")
test()
```

Gives:

```
[...]
-------------------- second run ---------------------

-------------------- batch --------------------------
  0.198542 seconds (1.30 M allocations: 49.313 MiB, 6.03% gc time)
-------------------- alternating --------------------
 25.296735 seconds (1.20 M allocations: 47.019 MiB, 0.05% gc time)
```

It was a little harder to isolate than I expected, querying `is_integer` or `has_lower_bound` do not  trigger the slow path (my first try did not even have an associated bound), but `lower_bound` triggers the slow path. Seems like the first two call `MOI.is_valid` and Gurobi implementation of it is a cache that does not update the model to query it, while `lower_bound` calls for a `MOI.get` to get the corresponding constraint and return a field value, this is not cached in Gurobi and calls a model update.

Seems like I owe you an apology, whatever the slowdown caused by broadcast instead of dedicated batch methods it is not relevant in this case (Gurobi), not in comparison with the slowdown caused by alternating commands that trigger `_require_update` and `_update_if_needed`.

As CPLEX does not have the same semantics (it implicitly does an update after each call, as its model is always up-to-date), I will keep my draft PR to JuMP open until I can test CPLEX in a more detailed manner, to check if in CPLEX case the batch versions of fix/unfix/etc are justifiable, or it is just simpler to stick with broadcast anyway. 

I understand what you say while, me being a pedant, I say that "there are always going to be work-flows that hit the slow path" is not exactly right, you can *cache every change* and so `update_model!` is only called in `optimize!`, but I understand that this is impracticable.

It is not the first time I am bitten by this behavior of the Gurobi wrapper, and it is a little harder to debug it when it is not every alternating sequence of commands that trigger it. What do you think about a subsection for the [Use with JuMP](https://github.com/JuliaOpt/Gurobi.jl/#use-with-jump) section of the README.md called something like 'Caveats' to remember the user that the `update` is controlled by the wrapper and strange slowdowns may be fixed by doing changes in broadcasted steps? If you think it is worth it I may write something for you to review.

odow:
One option is to cache the lower and upper bounds, and return them from the cache rather than asking Gurobi via C. That should alleviate the most common sources.

A PR to update the documentation is most welcome. The functionality is:

1. Gurobi stores modifications in a lazy buffer
2. These modifications are processed when one of two things occurs: (i) `optimize!` is called; (ii) when ANY model attribute is queried (e.g., a variable lower bound), **even if it has not changed since the last query**.
3. Modifying the model is fast, because modifications are pushed into the buffer
4. If the buffer is empty, querying attributes is fast because no changes have to be computed.
5. The worst-case scenario is modify-query-modify-query.
For example, instead of
```Julia
model = Model(Gurobi.Optimizer)
@variable(model, x[1:100] >= 0)
for i = 1:100
    set_upper_bound(x[i], i)
    println(lower_bound(x[I]))
end
```
use
```Julia
model = Model(Gurobi.Optimizer)
@variable(model, x[1:100] >= 0)
for i = 1:100
    set_upper_bound(x[i], i)
end
for i = 1:100
    println(lower_bound(x[I]))
end
```

odow:
Can this be closed then?

henriquebecker91:
Yes. I was just waiting #301, #302, and #303 to be closed to close this one.

PR #301, opened 2020-03-27
Adds section Common Performance Pitfall with JuMP.

henriquebecker91:
As discussed with @odow in https://github.com/JuliaOpt/Gurobi.jl/pull/300, this updates the documentation of README.md to give new JuMP.jl+Gurobi.jl users a warning about the most common performance pitfall (so they may be already aware of it and structure their JuMP programs accordingly). This will also probably help users that find the "excessive updates" message and search for how to fix it.


henriquebecker91:
> Thanks for your patience.

I could say the same.

I will look into the bounds cache idea, but no guarantees, if I fell it
would add a lot of code or risk messing with existing functionality then I
will not open a PR.

Em sex., 27 de mar. de 2020 às 18:52, Oscar Dowson <notifications@github.com>
escreveu:

PR #308, opened 2020-04-11
Don't hide the error being thrown by Gurobi when it fails to compute …

### LINKED ISSUE #307: Error when retrieving the conflict status of a feasible MIP model
Consider the following LP example:
```julia
using JuMP, Gurobi
m = Model(Gurobi.Optimizer)
@variable(m, 0 <= x <= 2)
@variable(m, 0 <= y <= 30)
@objective(m, Max, 1*x + 2*y)
@constraint(m, 1x + 5y <= 3.0)
optimize!(m)
```
It is feasible and I get: 
```julia
julia> termination_status(m)
OPTIMAL::TerminationStatusCode = 1

julia> Gurobi.compute_conflict(backend(m).optimizer.model)
IIS runtime: 0.00 seconds

julia> JuMP.MOI.get(m, Gurobi.ConflictStatus())
INFEASIBLE::TerminationStatusCode = 2
```
But if I have a MIP problem:
```julia
using JuMP
m = Model(Gurobi.Optimizer)
@variable(m, x, Bin)
@variable(m, 0 <= y <= 30)
@objective(m, Max, 1*x + 2*y)
@constraint(m, 1x + 5y <= 3.0)
optimize!(m)
```
I get something different:
```julia
julia> termination_status(m)
OPTIMAL::TerminationStatusCode = 1

julia> Gurobi.compute_conflict(backend(m).optimizer.model)
Found heuristic solution: objective -0.0000000

Explored 0 nodes (0 simplex iterations) in 0.00 seconds
Thread count was 1 (of 4 available processors)

Solution count 1: -0
No other solutions better than 0

Optimal solution found (tolerance 1.00e-04)
Best objective -0.000000000000e+00, best bound -0.000000000000e+00, gap 0.0000%
IIS runtime: 0.00 seconds

julia> JuMP.MOI.get(m, Gurobi.ConflictStatus())
ERROR: OptimizeNotCalled()
```

This might actually be a problem with `Gurobi.jl`.

odow:
…the IIS

We shouldn't be hiding this error, because Gurobi behaves differently with continuous and mixed-integer models. See #307 for details.

image

Closes #307.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -2578,15 +2578,7 @@ conflict is not purged, and any calls to the above attributes will return values
 for the original conflict without a warning.
 """
 function compute_conflict(model::Optimizer)
-    try
-        computeIIS(model.inner)
-    catch exc
-        if isa(exc, GurobiError) && exc.code == 10015
-            model.inner.conflict = Gurobi.GRB_INFEASIBLE
-        else
-            rethrow(exc)
-        end
-    end
+    computeIIS(model.inner)
     return
 end
 
--- a/test/MOI/MOI_wrapper.jl
+++ b/test/MOI/MOI_wrapper.jl
@@ -403,11 +403,12 @@ end
         @test MOI.get(model, Gurobi.ConflictStatus()) == MOI.OPTIMIZE_NOT_CALLED
         @test_throws ErrorException MOI.get(model, Gurobi.ConstraintConflictStatus(), c1)
 
-        # Once it's called, no problem.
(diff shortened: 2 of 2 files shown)


odow:
@ohmsweetohm1, thoughts?

ohmsweetohm1:
I like it! Now it is clearer:
`ERROR: LoadError: Gurobi.GurobiError(10015, "Cannot compute IIS on a feasible model")` in both the LP and the MIP case.

odow:
Great. We were trying to be too clever before.

PR #312, opened 2020-04-20
Revert QCPDual to default

### LINKED ISSUE #309: On QCQP in Gurobi v9.0
I noticed the following,
- `QCPDual` seems to be set to 1 by default, which differs from Gurobi's documented semantics
- When `QCPDual` is not set, the SolveTime attribute seems to return only the dual solve time and not the overall solve time for primal and dual.

odow:
Closes #309.

cc @ccoffrin

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Gurobi"
 uuid = "2e9cd046-0924-5485-92f1-d5272153d98b"
 repo = "https://github.com/JuliaOpt/Gurobi.jl"
-version = "0.7.7"
+version = "0.8.0"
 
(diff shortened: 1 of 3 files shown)


odow:
I guess this is technically a breaking change.

blegat:
Yes, we can do a minor version bump

odow:
If we're reverting `QCPDual`, we should also revert `InfUnbdInfo` and break multiple things at the same time:
image

ccoffrin:
if "revert `InfUnbdInfo`" means going back to Gurobi's defaults, I agree!

PR #313, opened 2020-04-20
Adds a specialization of MOI.delete that deletes a batch of constraints.

henriquebecker91:
I have only did it for MOI.ScalarAffineFunction constraints. This is merely a performance improvement. JuMP and MathOptInterface already have batch constraint deletion fallbacks and they will correctly call this method when it is the case. I did not add new tests nor benchmarks, but did call `JuMP.delete` in my code (that had performance problems), checked if the method was really called (with a `println` already removed), and the performance difference was from 23.2 seconds to 0.531 seconds (~29k constraints deleted from ~36k constraints, the speedup was probably higher than the ~43x because I am timing two batch variable deletions in these measurements, instead of just the constraint deletion, and they did not change performance with the changes introduced by this commit).

I executed the Gurobi tests as usual:
[](https://postimg.cc/PLZMftsv)

@odow can you give it a look? If you want me to stop pinging you in these cases just tell me, I am never sure if I should or not. If you think it needs extra tests or a "minimal working benchmark" I can work on them asap.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -1469,6 +1469,29 @@ end
 
 function MOI.delete(
     model::Optimizer,
+    cs::Vector{<:MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64}, <:Any}}
+)
+    # The `sort!`s called are necessary for improving the efficiency of the
+    # updates (i.e., without them we cannot use `searchsorted*` methods).
+    _update_if_necessary(model)
+    rows_to_delete = sort!(getfield.(_info.(model, cs), :row))
+    del_constrs!(model.inner, rows_to_delete)
+    _require_update(model)
+    for (_, info) in model.affine_constraint_info
+        info.row -= searchsortedlast(rows_to_delete, info.row)
+    end
+    cs_values = sort!(getfield.(cs, :value))
+    # If the key of an model.affine_constraint_info entry is in cs_values,
+    # then that entry is deleted.
+    filter!(model.affine_constraint_info) do pair
+        isempty(searchsorted(cs_values, pair.first))
+    end
+    model.name_to_constraint_index = nothing
+    return
+end
+
+function MOI.delete(
+    model::Optimizer,
     c::MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64}, <:Any}
 )
     row = _info(model, c).row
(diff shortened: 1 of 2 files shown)


odow:
You are welcome to ping me :)

Does this get called in the MOI tests? At minimum, we need to ensure that added code is called. This can either be in the MOI tests, or via a specific test in Gurobi.jl.

src/MOI_wrapper.jl
```diff
@@ -1467,6 +1467,29 @@ function MOI.add_constraints(
...
+    cs::Vector{<:MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64}, <:Any}}
+)
+    # The `sort!`s called are necessary for improving the efficiency of the
+    # updates (i.e., without them we cannot use `searchsorted*` methods).
+    _update_if_necessary(model)
+    rows_to_delete = sort!(getfield.(_info.(model, cs), :row))
+    del_constrs!(model.inner, rows_to_delete)
+    _require_update(model)
+    for (_, info) in model.affine_constraint_info
+        info.row -= searchsortedlast(rows_to_delete, info.row)
```
blegat, line 1481: I would add a comment explaining what this loop does. IIUC, for alll constraints, it gets the index of the last row with row index smaller. That gives the number of rows before, hence substracting this give the new row index

henriquebecker91: I will do it. I thought about writing this comment already. However I have already used this trick in other places (I think in both Gurobi.jl and CPLEX.jl) and did not comment it. Should I add it on all these places?

henriquebecker91:
This is a screenshot of the `Gurobi.jl` tests after the last commit but with a `println` added in the code to be extra sure the method is called. 

[](https://postimg.cc/w7H4dMWn)

henriquebecker91:
Thanks for the fast response!

src/MOI_wrapper.jl
blegat: Yes, it would help there also IMO

henriquebecker91: Created a PR for [Gurobi.jl](https://github.com/JuliaOpt/Gurobi.jl/pull/314) and another for [CPLEX.jl](https://github.com/JuliaOpt/CPLEX.jl/pull/298).

PR #317, opened 2020-05-04
Adds buffering for variables deleted.

### LINKED ISSUE #315: More fun with lazy update semantics
Raised on Discourse: https://discourse.julialang.org/t/jump-returning-incorrect-value-after-deleting-adding-anonymous-variable-and-constraint/38036

I don't know whether this is a bug, or expected behavior. I coded in Gurobipy and can't reproduce.

```Julia
using Gurobi

function run()
    env = Gurobi.Env()
    model = Gurobi.Model(env, "")
    Gurobi.add_var!(model, Cchar('C'), 0.0, -Inf, Inf)
    Gurobi.del_vars!(model, Cint[1])
    Gurobi.add_var!(model, Cchar('C'), 0.0, -Inf, Inf)
    Gurobi.set_dblattrelement!(model, "LB", 1, 1.0)
    Gurobi.update_model!(model)
    return Gurobi.get_dblattrelement(model, "LB", 1)
end
run()  # Returns -1e100 instead of 1.
```

henriquebecker91:
Gurobi.Optimizer now have fields `next_column` and `columns_deleted_since_last_update`. The behavior now is to not force a update before deleting, but instead: update the list of variables in Julia side but not their column indexes, and save the column indexes of the ones deleted; when updating the model then update the column indexes in Julia side.

The code pass all tests. But maybe we need to devise new ones.

This fix #315 in my code (that unfortunately is not a MWE).

I will be out for about 2 hours now, @odow, after this I will check if you have suggested something to me to improve in this PR.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -87,6 +87,24 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     # variable type, and the names of SingleVariable-in-Set constraints.
     variable_info::CleverDicts.CleverDict{MOI.VariableIndex, VariableInfo}
 
+    # If you add variables to a model that had variables deleted AND has
+    # not called `update_model!` since the deletion, then the newly created
+    # variables may have attributes set, but their column index before the
+    # call to `update_model!` is different than after the `update_model!`.
+    # Before the `update_model!` their column is the same as if no variables
+    # were deleted, after the `update_model!` the columns indexes are
+    # shifted (by being being subtracted by the number of variables deleted
+    # with column indexes smaller than them). To control this the two
+    # fields below are used:
+    # `next_column`: The column index of the next variable/column added. It is
+    # updated when variables are added, and when the `_update_if_necessary!` is
+    # called AND `columns_deleted_since_last_update` is not empty.
+    # `columns_deleted_since_last_update`: Stores the column indexes of all
+    # columns that were deleted since the last call to `_update_if_necessary!`,
+    # after such call the vector is emptied.
+    next_column::Int
+    columns_deleted_since_last_update::Vector{Int}
+
     # An index that is incremented for each new constraint (regardless of type).
     # We can check if a constraint is valid by checking if it is in the correct
     # xxx_constraint_info. We should _not_ reset this to zero, since then new
@@ -135,6 +153,8 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
         model.silent = false
         model.params = Dict{String, Any}()
         model.variable_info = CleverDicts.CleverDict{MOI.VariableIndex, VariableInfo}()
+        model.next_column = 1
+        model.columns_deleted_since_last_update = Int[]
         model.affine_constraint_info = Dict{Int, ConstraintInfo}()
         model.quadratic_constraint_info = Dict{Int, ConstraintInfo}()
         model.sos_constraint_info = Dict{Int, ConstraintInfo}()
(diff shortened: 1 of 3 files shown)


src/MOI_wrapper.jl
```diff
@@ -419,6 +453,12 @@ function _info(model::Optimizer, key::MOI.VariableIndex)
     throw(MOI.InvalidIndex(key))
 end
 
+function _get_next_column(model::Optimizer)
+    old_value = deepcopy(model.next_column)
```
odow, line 457: Why is `deepcopy` needed?

src/MOI_wrapper.jl
```diff
@@ -419,6 +453,12 @@ function _info(model::Optimizer, key::MOI.VariableIndex)
...
+    model.next_column += 1
+    return old_value
```
odow, line 459:
```Julia
function _get_next_column(model::Optimizer)
    model.next_column += 1
    return model.next_column - 1
end
```

henriquebecker91:
`deepcopy` is not needed per se, is just an habit of mine to not assume
fields will not be changed to mutable types after. It should be compiled to
a no-operation, but I can remove it if you think it is better.

I will make your example a test then. Just not now.

Em seg., 4 de mai. de 2020 às 16:13, Oscar Dowson <notifications@github.com>
escreveu:

henriquebecker91:
@odow The test was a little harder to create than I expected. Many common attributes are cached on the wrapper (name, bounds, type, ...) and other are discarded (Start, VBasis, ...) if any variables are deleted. This explains how this bug did not surface immediately, and how the code seems to work in some situations that it shouldn't work, given the nature of the bug.

The testset has 4 fails in current master branch and none with the changes introduced by this branch. I think it is a solid test.

src/MOI_wrapper.jl
henriquebecker91: Done.

odow:
Which are the tests that fail?

henriquebecker91:
```
Buffered deletion test: Test Failed at /home/henrique/AreaDeTrabalho/Gurobi.jl/test/MOI/MOI_wrapper.jl:519
  Expression: vars_obj[2] == MOI.get(model, obj_attr, vars[2])
   Evaluated: 42.0 == -77.0
Buffered deletion test: Test Failed at /home/henrique/AreaDeTrabalho/Gurobi.jl/test/MOI/MOI_wrapper.jl:521
  Expression: fourth_var_obj == MOI.get(model, obj_attr, fourth_var)
   Evaluated: -77.0 == 0.0
Buffered deletion test: Test Failed at /home/henrique/AreaDeTrabalho/Gurobi.jl/test/MOI/MOI_wrapper.jl:526
  Expression: vars_obj[2] == MOI.get(model, obj_attr, vars[2])
   Evaluated: 42.0 == -77.0
Buffered deletion test: Test Failed at /home/henrique/AreaDeTrabalho/Gurobi.jl/test/MOI/MOI_wrapper.jl:528
  Expression: fourth_var_obj == MOI.get(model, obj_attr, fourth_var)
   Evaluated: -77.0 == 0.0
```

Basically the tests that check the value of the property set. The ones before the implicit `update_model!` caused by `MOI.optimize!` and the ones after it. In the patched version such errors do not exist (all tests pass). So I am confident the code, while not bug-free, solves the bug it intends to solve and do not break any other test.

odow:
Oops. I mis-read your comment. I understood it as 4 different tests were failing, but these ones were passing.

henriquebecker91:
Oh, no. With this PR all older tests pass and the new ones too. In the master branch, the 4 `@test` I cited of the new `@testset` I created fail (because the bug the PR solves).

henriquebecker91:
@odow, do you want me to improve this PR in some way? While this is not approved, Gurobi.jl will cause subtle bugs to users when deletion is followed by attribute changes. It is also good for me if I can refer to the Gurobi version I used in my paper, instead of saying I used a patched local copy, XD.

odow:
Sorry, merging since this is a strict improvement in correcting an existing bug. Thanks!

PR #326, opened 2020-05-20
Update IIS support to MOI 0.9.14.

dourouc05:
JuliaOpt/MathOptInterface.jl#1056 is merged and will be included in the next MOI release.

Several return codes are not yet in MOI, though. Should they be included in MOI?

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -2663,7 +2663,7 @@ computed conflict. It returns an integer:
 
 * `-1` if `compute_conflict!` has not yet been called
 * `0` if it found a conflict
-* other values are defined in [Gurobi's documentation](https://www.gurobi.com/documentation/9.0/refman/error_codes.html)
+* other values are defined in [Gurobi's documentation](https://www.gurobi.com/documentation/current/refman/error_codes.html)
 """
 struct ConflictStatus <: MOI.AbstractModelAttribute end
 
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -2655,12 +2655,29 @@ function _is_feasible(model::Optimizer)
     return model.inner.conflict == Gurobi.GRB_INFEASIBLE
 end
 
+"""
+    ConflictStatus()
+
+Return the raw status from Gurobi indicating the status of the last
+computed conflict. It returns an integer:
+
+* `-1` if `compute_conflict!` has not yet been called
+* `0` if it found a conflict
+* other values are defined in [Gurobi's documentation](https://www.gurobi.com/documentation/9.0/refman/error_codes.html)
+"""
+struct ConflictStatus <: MOI.AbstractModelAttribute end
+
+function MOI.get(model::Optimizer, ::ConflictStatus)
+    return model.inner.conflict
+end
+
 function MOI.get(model::Optimizer, ::MOI.ConflictStatus)
-    if model.inner.conflict == -1
+    status = MOI.get(model, ConflictStatus())
+    if status == -1
         return MOI.COMPUTE_CONFLICT_NOT_CALLED
-    elseif model.inner.conflict == 0
+    elseif status == 0
         return MOI.CONFLICT_FOUND
-    elseif model.inner.conflict == Gurobi.IIS_NOT_INFEASIBLE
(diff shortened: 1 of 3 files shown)


src/MOI_wrapper.jl
```diff
@@ -2640,25 +2628,14 @@ function _is_feasible(model::Optimizer)
...
+        return MOI.COMPUTE_CONFLICT_NOT_CALLED
     elseif model.inner.conflict == 0
-        return MOI.OPTIMAL
-    elseif model.inner.conflict == Gurobi.GRB_LOADED
+        return MOI.CONFLICT_FOUND
+    elseif model.inner.conflict == Gurobi.IIS_NOT_INFEASIBLE
+        return MOI.NO_CONFLICT_EXISTS
+    elseif model.inner.conflict == Gurobi.GRB_LOADED # TODO: these codes do not exist in MOI.
         return MOI.OTHER_ERROR
     elseif model.inner.conflict == Gurobi.GRB_OPTIMAL
```
odow, line 2640: See my comment on CPLEX.

dourouc05:
Based on JuliaOpt/MathOptInterface.jl#1091, I used `NO_CONFLICT_FOUND` when Gurobi has aborted.

Is it good enough to be merged? (Once MOI 0.9.14 is tagged.)

src/MOI_wrapper.jl
```diff
@@ -2612,19 +2612,7 @@ function MOI.get(
...
-infeasible.
-
-See also `Gurobi.ConflictStatus` and `Gurobi.ConstraintConflictStatus`.
-
-Note that if `model` is modified after a call to `compute_conflict`, the
-conflict is not purged, and any calls to the above attributes will return values
-for the original conflict without a warning.
-"""
-function compute_conflict(model::Optimizer)
+function MOI.compute_conflict!(model::Optimizer)
```
blegat, line 2615: Could you add `@deprecate compute_conflict MOI.compute_conflict!` ?

src/MOI_wrapper.jl
```diff
@@ -2640,129 +2628,96 @@ function _is_feasible(model::Optimizer)
...
-"""
-    ConflictStatus()
-
-Return an `MOI.TerminationStatusCode` indicating the status of the last
-computed conflict. If a minimal conflict is found, it will return
-`MOI.OPTIMAL`. If the problem is feasible, it will return `MOI.INFEASIBLE`. If
-`compute_conflict` has not been called yet, it will return
-`MOI.OPTIMIZE_NOT_CALLED`.
-"""
-struct ConflictStatus <: MOI.AbstractModelAttribute end
```
blegat, line 2652: Could you add `@deprecate ConflictStatus MOI.ConflictStatus` ?

blegat: Even better would be to keep `Gurobi.ConflictStatus` and let it return the solver-specific `Gurobi.GRB_...` and call this from the getter of `MOI.ConflictStatus`. That would allow the user to have the solver-specific status in case he wants more details.

dourouc05: Isn't defining the same attribute for several solvers defeating the purpose of MOI? Wouldn't it be better to have something like `MOI.RawConflictStatus`?

blegat: The purpose of MOI is to have solver-independent and solver-dependent features coexist seemlessly.
Here, it's not more work to add a Gurobi-specific attribute that could be accessed even at the JuMP level.
We could have a `RawConflictStatus` but before we do, we could have solver-specific attributes.
The situation is similar for callbacks, we have both solver-independent ones and solver-dependent ones.

dourouc05: Here it is (with tests included)!

dourouc05:
With MOI 0.9.14 being tagged, maybe this could get merged?

odow:
Oops. Sorry for dropping the ball on this. Feel free to ping earlier :) Can you post a screenshot with the tests passing? Looks good otherwise.

dourouc05:
Same as for CPLEX: 

image
image
image

blegat:
Can you rebase to solve the conflict

dourouc05:
Done.

PR #350, opened 2020-09-29
Refactor tests to pass with limited licenses

### LINKED ISSUE #336: Tests Fail with Limited Licenses
I tried running the tests with 3 floating licenses.  All three were consumed and then tests started to fail.  Tests should be updated to use at most 1 license.

odow:
Closes #336 

@ccoffrin want to test?


odow:
That is the correct output. I think it needs 4 to run all the tests, because we explicitly test having multiple models open at the same time with different environments.

PR #363, opened 2020-10-28
Add farkas dual for variable bounds

odow:
cc @juan-pablo-vielma

image

TODO:
- [x] Are the signs correct for min/max?
- [x] Are the signs correct for `EqualTo` and `Interval`?

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -2525,17 +2525,20 @@ The Farkas dual of the variable is ā, and it applies to the upper bound if ā
 and it applies to the lower bound if ā > 0.
 """
 function _farkas_variable_dual(model::Optimizer, col::Cint)
-    valueP = Ref{Cint}()
-    ret = GRBgetintattr(model, "NumConstrs", valueP)
+    numnzP = Ref{Cint}()
+    ret = GRBgetvars(model, numnzP, C_NULL, C_NULL, C_NULL, col, 1)
+    _check_ret(model, ret)
+    vbeg = Vector{Cint}(undef, 2)
+    vind = Vector{Cint}(undef, numnzP[])
+    vval = Vector{Cdouble}(undef, numnzP[])
+    ret = GRBgetvars(model, numnzP, vbeg, vind, vval, col, 1)
     _check_ret(model, ret)
     ā = 0.0
-    λ, a = Ref{Cdouble}(), Ref{Cdouble}()
-    for row = 1:valueP[]
-        ret = GRBgetdblattrelement(model, "FarkasDual", Cint(row - 1), λ)
+    λ = Ref{Cdouble}()
+    for (ind, val) in zip(vind, vval)
+        ret = GRBgetdblattrelement(model, "FarkasDual", ind, λ)
         _check_ret(model, ret)
-        ret = GRBgetcoeff(model, Cint(row - 1), col, a)
-        _check_ret(model, ret)
-        ā += λ[] * a[]
+        ā += λ[] * val
     end
     return ā
 end
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -2546,14 +2546,13 @@ function MOI.get(
 )
     _throw_if_optimize_in_progress(model, attr)
     MOI.check_result_index_bounds(model, attr)
+    col = Cint(column(model, c) - 1)
     if model.has_infeasibility_cert
-        dual = _farkas_variable_dual(model, Cint(column(model, c) - 1))
+        dual = _farkas_variable_dual(model, col)
         return min(dual, 0.0)
     end
     reduced_cost = Ref{Cdouble}()
-    ret = GRBgetdblattrelement(
-        model, "RC", Cint(column(model, c) - 1), reduced_cost
-    )
+    ret = GRBgetdblattrelement(model, "RC", col, reduced_cost)
     _check_ret(model, ret)
     sense = MOI.get(model, MOI.ObjectiveSense())
     # The following is a heuristic for determining whether the reduced cost
@@ -2581,14 +2580,13 @@ function MOI.get(
 )
     _throw_if_optimize_in_progress(model, attr)
     MOI.check_result_index_bounds(model, attr)
+    col = Cint(column(model, c) - 1)
     if model.has_infeasibility_cert
-        dual = _farkas_variable_dual(model, Cint(column(model, c) - 1))
+        dual = _farkas_variable_dual(model, col)
         return max(dual, 0.0)
     end
     reduced_cost = Ref{Cdouble}()
-    ret = GRBgetdblattrelement(
-        model, "RC", Cint(column(model, c) - 1), reduced_cost
-    )
+    ret = GRBgetdblattrelement(model, "RC", col, reduced_cost)
     _check_ret(model, ret)
     sense = MOI.get(model, MOI.ObjectiveSense())
     # The following is a heuristic for determining whether the reduced cost
@@ -2616,13 +2614,12 @@ function MOI.get(
 )
     _throw_if_optimize_in_progress(model, attr)
     MOI.check_result_index_bounds(model, attr)
+    col = Cint(column(model, c) - 1)
     if model.has_infeasibility_cert
-        return _farkas_variable_dual(model, Cint(column(model, c) - 1))
+        return _farkas_variable_dual(model, col)
     end
     reduced_cost = Ref{Cdouble}()
-    ret = GRBgetdblattrelement(
-        model, "RC", Cint(column(model, c) - 1), reduced_cost
-    )
+    ret = GRBgetdblattrelement(model, "RC", col, reduced_cost)
     _check_ret(model, ret)
     return _dual_multiplier(model) * reduced_cost[]
 end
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -2512,13 +2512,45 @@ function _dual_multiplier(model::Optimizer)
...
+function _farkas_variable_dual(model::Optimizer, col::Cint)
+    valueP = Ref{Cint}()
+    ret = GRBgetintattr(model, "NumConstrs", valueP)
+    _check_ret(model, ret)
+    ā = 0.0
+    λ, a = Ref{Cdouble}(), Ref{Cdouble}()
+    for row = 1:valueP[]
+        ret = GRBgetdblattrelement(model, "FarkasDual", Cint(row - 1), λ)
+        _check_ret(model, ret)
+        ret = GRBgetcoeff(model, Cint(row - 1), col, a)
```
mlubin, line 2536: Looping through all entries in the constraint matrix is *way* too expensive to do by default.

odow: Ah, I missed this: https://www.gurobi.com/documentation/9.0/refman/c_getvars.html

juan-pablo-vielma: using `GRBgetvars` should be faster.

juan-pablo-vielma:
the `GRBgetvars` version works on my example too.

src/MOI_wrapper.jl
```diff
@@ -2512,13 +2512,48 @@ function _dual_multiplier(model::Optimizer)
...
+    _check_ret(model, ret)
+    vbeg = Vector{Cint}(undef, 2)
+    vind = Vector{Cint}(undef, numnzP[])
+    vval = Vector{Cdouble}(undef, numnzP[])
+    ret = GRBgetvars(model, numnzP, vbeg, vind, vval, col, 1)
+    _check_ret(model, ret)
+    ā = 0.0
+    λ = Ref{Cdouble}()
+    for (ind, val) in zip(vind, vval)
+        ret = GRBgetdblattrelement(model, "FarkasDual", ind, λ)
```
joaquimg, line 2539: You can probably use the `GRBgetdblattrlist` version

test/MOI/MOI_wrapper.jl
```diff
@@ -1177,6 +1173,58 @@ function test_indicator_nonconstant()
...
+        MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([2.0, 1.0], x), 0.0),
+        MOI.LessThan(-1.0),
+    )
+    MOI.optimize!(model)
+    @test MOI.get(model, MOI.TerminationStatus()) == MOI.INFEASIBLE
+    @test MOI.get(model, MOI.DualStatus()) == MOI.INFEASIBILITY_CERTIFICATE
+    clb_dual = MOI.get.(model, MOI.ConstraintDual(), clb)
+    c_dual = MOI.get(model, MOI.ConstraintDual(), c)
+    @test clb_dual[1] ≈ -2 * c_dual atol = 1e-6
+    @test clb_dual[2] ≈ -c_dual atol = 1e-6
```
odow, line 1225: @blegat, it's not obvious to me what our sign conventions are.

Currently for `MIN_SENSE`, Gurobi returns
```Julia
(clb_dual, c_dual) = ([2.0, 1.0], -1.0)
```
and for `MAX_SENSE`, it returns
```julia
(clb_dual, c_dual) = ([-1.0, -0.5], 0.5)
```
Should the `MAX_SENSE` values be negated?

odow: I've negated the `MAX_SENSE` case. The latest commit returns `(clb_dual, c_dual) = ([1.0, 0.5], -0.5)`. I think this makes sense from our definition of duality, but I'm not sure it is what users will expect.

odow:
image

test/MOI/MOI_wrapper.jl
odow: Hmm. This might be a bug in Gurobi.
```Julia
julia> function test_farkas(c1)
           env = Ref{Ptr{Cvoid}}()
           ret = GRBloadenv(env, C_NULL)
           GRBsetintparam(env[], "InfUnbdInfo", 1)
           GRBsetintparam(env[], "OutputFlag", 0)
           model = Ref{Ptr{Cvoid}}()
           GRBloadmodel(
               env[],  # *env
               model,  # **modelp
               C_NULL,  # *Pname
               Cint(2),  # numvars
               Cint(1),  # numconstrs
               Cint(-1),  # objsense (maximization)
               0.0, # objcon
               Cdouble[c1, 0.0],  # *obj
               Cchar[GRB_LESS_EQUAL],  # *sense
               Cdouble[-1.0], # *rhs
               Cint[0, 1], # *vbeg
               Cint[1, 1],  # vlen
               Cint[0, 0],  # *vind
               Cdouble[-2.0, -1.0],  # *vval
               Cdouble[-GRB_INFINITY, -GRB_INFINITY],  # *lb
               Cdouble[0.0, 0.0],  # *ub
               C_NULL,  # *vtype
               C_NULL,  # **varnames
               C_NULL,  # **constrnames
           )
           GRBoptimize(model[])
           dual = Vector{Cdouble}(undef, 1)
           GRBgetdblattrarray(model[], "FarkasDual", 0, 1, dual)
           dual
       end
test_farkas (generic function with 1 method)

julia> test_farkas(1.0)
1-element Array{Float64,1}:
 1.0

julia> test_farkas(-1.0)
1-element Array{Float64,1}:
 -0.5
```

Would we expect the Farkas multiplier to switch based on the sign of an objective coefficient?

It seems to contradict their documentation:
https://www.gurobi.com/documentation/9.0/refman/farkasdual.html#attr:FarkasDual

image

odow: Seems to happen only with upper-bounded variables, so this might be a presolve transform leaking through. I've put in a support request.

blegat: Isn't it `@test clb_dual[2] ≈ -c_dual - 1` to account for the objective?

blegat:
That that you could use `MOI.Utilities.get_fallback` to get the duals of variable bounds computed from other duals.

test/MOI/MOI_wrapper.jl
odow: I guess it depends if you do `A' y >= 0`, or `A' y >= b` (i.e., Farkas, or a unbounded ray). 

You previously suggested it was just `club_dual[2] == -c_dual`:
https://github.com/jump-dev/MathOptInterface.jl/pull/127#issuecomment-329997053

The documentation for `INFEASIBILITY_CERTIFICATE` just says that it is a proof of infeasibility, but doesn't specify more:
https://github.com/jump-dev/MathOptInterface.jl/blob/7543afe4b5151cf36bbd18181c1bb5c83266ae2f/src/attributes.jl#L1339-L1340

blegat: Indeed, I missed that it was infeasible

test/MOI/MOI_wrapper.jl
```diff
@@ -1177,6 +1179,196 @@ function test_indicator_nonconstant()
...
+    @test MOI.get(model, MOI.TerminationStatus()) == MOI.INFEASIBLE
+    @test MOI.get(model, MOI.DualStatus()) == MOI.INFEASIBILITY_CERTIFICATE
+    clb_dual = MOI.get.(model, MOI.ConstraintDual(), clb)
+    c_dual = MOI.get(model, MOI.ConstraintDual(), c)
+    @show clb_dual, c_dual
+    @test clb_dual[1] < 1e-6
+    @test clb_dual[2] < 1e-6
+    @test_broken c_dual[1] < 1e-6
+    @test_broken clb_dual[1] ≈ 2 * c_dual atol = 1e-6
+    @test_broken clb_dual[2] ≈ c_dual atol = 1e-6
```
odow, line 1369: TODO: Gurobi, including 9.1 is broken on these tests. I have a support request in.

odow: Here's an example demonstrating this is clearly a bug. The sign switches at a lower bound of `-1e20`, which is the value at which Gurobi ignores the lower bound.
```Julia
function test_farkas(obj_coefficient, lower_bound)
    env = Ref{Ptr{Cvoid}}()
    ret = GRBloadenv(env, C_NULL)
    GRBsetintparam(env[], "InfUnbdInfo", 1)
    GRBsetintparam(env[], "OutputFlag", 0)
    model = Ref{Ptr{Cvoid}}()
    GRBloadmodel(
        env[],                      # *env
        model,                      # **modelp
        C_NULL,                     # *Pname
        Cint(1),                    # numvars
        Cint(1),                    # numconstrs
        Cint(1),                    # objsense
        0.0,                        # objcon
        Cdouble[obj_coefficient],   # *obj
        Cchar[GRB_GREATER_EQUAL],   # *sense
        Cdouble[1.0],               # *rhs
        Cint[0],                    # *vbeg
        Cint[1],                    # vlen
        Cint[0],                    # *vind
        Cdouble[1.0],               # *vval
        Cdouble[lower_bound],       # *lb
        Cdouble[0.0],               # *ub
        C_NULL,                     # *vtype
        C_NULL,                     # **varnames
        C_NULL,                     # **constrnames
    )
    GRBoptimize(model[])
    dual = Vector{Cdouble}(undef, 1)
    GRBgetdblattrarray(model[], "FarkasDual", 0, 1, dual)
    return dual[1]
end

julia> test_farkas(1.0, -1e20)
Academic license - for non-commercial use only - expires 2020-12-05
1.0

julia> test_farkas(1.0, nextfloat(-1e20))
Academic license - for non-commercial use only - expires 2020-12-05
-1.0

julia> a, b, c = Ref{Cint}(), Ref{Cint}(), Ref{Cint}()
(Base.RefValue{Int32}(482535088), Base.RefValue{Int32}(482535088), Base.RefValue{Int32}(482535152))

julia> GRBversion(a, b, c)

julia> a, b, c
(Base.RefValue{Int32}(9), Base.RefValue{Int32}(1), Base.RefValue{Int32}(0))
```

odow: Still waiting for Gurobi support, but I'm now fairly confident that this is a bug in Gurobi, and therefore not blocking for this PR. 

If they fix it in a future release, the `@test_broken` will break (pass) and we can address.

odow: Bug confirmed upstream. Will be fixed in next release.

PR #380, opened 2020-12-21
MOI.set for basis status

joehuchette:
I've added two methods to ``MOI.set`` that allow you to set variable/constraint basis status. This is surprisingly difficult behavior to test: I don't think it's possible to check the basis status in Gurobi without optimizing the model, at which point you can only indirectly measure if your basis was correctly set.

I also looked at adding tests at the MOI level, but it got complicated when I tried to figure out how to make this work with universal fallback, so I punted.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -3181,6 +3181,60 @@ function MOI.get(
     end
 end
 
+function MOI.set(
+    model::Optimizer,
+    ::MOI.ConstraintBasisStatus,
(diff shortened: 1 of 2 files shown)


joehuchette:
Gurobi said that to unit test this, we should set an iteration limit (close to what I do here) or use a message callback: https://support.gurobi.com/hc/en-us/community/posts/360075729911-Getting-VBasis-CBasis-attributes-without-optimizing

odow:
Can you post a screenshot of the tests passing?

joehuchette:
image

PR #384, opened 2021-01-04
Add batch delete for SOS constraints and related test.

metab0t:
I have encountered performance issues in my code when a lot of SOS constraints are deleted from the model, so I add batch deletion of SOS constraints (based on existing batch deletion of affine constraints) and accompanying test.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -2218,6 +2218,27 @@ function MOI.delete(
     return
 end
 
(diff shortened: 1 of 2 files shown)


odow:
We don't have a CI setup for Gurobi.jl due to licensing reasons. Can you post a screenshot of the tests passing?

metab0t:
This is the test result. @odow 
Snipaste_2021-01-05_09-55-15

PR #385, opened 2021-01-05
Make Optimizer.variable_info field a concrete type

joehuchette:
Before, `Optimizer.variable_info::CleverDicts.CleverDict{MOI.VariableIndex,_VariableInfo}` was not a concrete type since the last two type parameters (for the hash and reverse_hash functions) were omitted. This lead to type instability, most notably for me in `_update_if_necessary`. Here's a pprof snippet from a larger program that solves a number of relatively easy LPs:

Screen Shot 2021-01-05 at 9 48 40 AM

This change 1) specifies a concrete type for `variable_info` and 2) changes the constructor called to make this work. The hash/reverse hash functions are identical to what would be used anyway by the originally used constructor.

Pprof after the change:

Screen Shot 2021-01-05 at 9 52 42 AM

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -106,6 +106,9 @@ end
 Base.cconvert(::Type{Ptr{Cvoid}}, x::Env) = x
 Base.unsafe_convert(::Type{Ptr{Cvoid}}, env::Env) = env.ptr_env::Ptr{Cvoid}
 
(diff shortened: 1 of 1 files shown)


odow (approved):
The extra type parameters were added in a recent MOI release.

I assume the tests still pass?

joehuchette:
Yes:

Screen Shot 2021-01-05 at 8 14 40 PM

PR #386, opened 2021-01-05
Add method to MOI.add_constrained_variable

joehuchette:
While doing some benchmarking, I noticed that calling `MOI.add_constrained_variable(::Gurobi.Optimizer, set::Gurobi._SCALAR_SET)` was spending roughly the same amount of time in `GRBaddvar` as it was in setting the bound attributes. This seemed a bit wasteful, so I added a new method that lives here that sets the bounds directly in `GRBaddvar`.

I'm a bit uncertain how to update `lower_bound_if_bounded` and `upper_bound_if_bounded`. The comments for `_VariableInfo` suggest that these fields should be NaN if and only if a variable has no bound constraints on it. That will never be the case in this method, so therefore I should set both bound values, even to +/-Inf for one-sided bounds. However, after looking through the rest of the file I'm not certain that this is the behavior actually implemented in, e.g., 
```
MOI.set(::Optimizer, ::MOI.ConstraintSet, ::MOI.ConstraintIndex{MOI.SingleVariable, S}, s::S) where {S <: Gurobi._SCALAR_SETS}
```
So, guidance would be appreciated on what I should do there.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -709,9 +709,14 @@ function MOI.add_variables(model::Optimizer, N::Int)
     return indices
 end
 
+# We implement a specialized version here to avoid calling into Gurobi twice.
+# Using the standard implementation, we would first create a variable in Gurobi
+# with GRBaddvar that has bounds of (-Inf,+Inf), and then immediately after
+# reset those bounds using the attributes interface. Instead, we just pass the
+# desired bounds directly to GRBaddvar.
 function MOI.add_constrained_variable(
     model::Optimizer, set::S
-) where {S <: _SCALAR_SETS}
+)::Tuple{MOI.VariableIndex,MOI.ConstraintIndex{MOI.SingleVariable, S}} where {S <: _SCALAR_SETS}
     vi = CleverDicts.add_item(
         model.variable_info, _VariableInfo(MOI.VariableIndex(0), 0)
     )
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -709,6 +709,47 @@ function MOI.add_variables(model::Optimizer, N::Int)
     return indices
 end
 
+function MOI.add_constrained_variable(
+    model::Optimizer, set::S
+) where {S <: _SCALAR_SETS}
+    vi = CleverDicts.add_item(
+        model.variable_info, _VariableInfo(MOI.VariableIndex(0), 0)
+    )
+    info = _info(model, vi)
+    # Now, set `.index` and `.column`.
+    info.index = vi
+    info.column = _get_next_column(model)
+    lb = -Inf
+    ub = Inf
+    if S <: MOI.LessThan{Float64}
+        info.bound = _LESS_THAN
+        ub = set.upper
+    elseif S <: MOI.GreaterThan{Float64}
+        info.bound = _GREATER_THAN
+        lb = set.lower
+    elseif S <: MOI.EqualTo{Float64}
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -709,6 +709,43 @@ function MOI.add_variables(model::Optimizer, N::Int)
...
+    )
+    info = _info(model, vi)
+    # Now, set `.index` and `.column`.
+    info.index = vi
+    info.column = _get_next_column(model)
+    lb = -Inf
+    ub = Inf
+    if S <: MOI.LessThan{Float64}
+        info.bound = _LESS_THAN
+        ub = set.upper
```
odow, line 726: Add a line `info.upper_bound_if_bounded = ub`

src/MOI_wrapper.jl
```diff
@@ -709,6 +709,43 @@ function MOI.add_variables(model::Optimizer, N::Int)
...
+    elseif S <: MOI.GreaterThan{Float64}
+        info.bound = _GREATER_THAN
+        lb = set.lower
```
odow, line 729: Add a line Add a line `info.lower_bound_if_bounded = lb`

src/MOI_wrapper.jl
```diff
@@ -709,6 +709,43 @@ function MOI.add_variables(model::Optimizer, N::Int)
...
+    elseif S <: MOI.EqualTo{Float64}
+        info.bound = _EQUAL_TO
+        lb = set.value
+        ub = set.value
```
odow, line 733: Add lines 
```julia
info.lower_bound_if_bounded = lb
info.upper_bound_if_bounded = ub
```

src/MOI_wrapper.jl
```diff
@@ -709,6 +709,43 @@ function MOI.add_variables(model::Optimizer, N::Int)
...
+    else
+        @assert S <: MOI.Interval{Float64}
+        info.bound = _INTERVAL
+        lb = set.lower
+        ub = set.upper
```
odow, line 738: Add lines 
```julia
info.lower_bound_if_bounded = lb
info.upper_bound_if_bounded = ub
```

src/MOI_wrapper.jl
```diff
@@ -709,6 +709,43 @@ function MOI.add_variables(model::Optimizer, N::Int)
...
+    end
+    info.lower_bound_if_bounded = lb
+    info.upper_bound_if_bounded = ub
```
odow, line 741: Remove these two lines

odow (commented):
The whole `_bound_if_bounded` thing was a hack to work around SecondOrderCones, because when adding a SOC, we might change the bound of the epigraph variable to 0, so it's not sufficient to just query Gurobi's internal value of the bound.

A better way, going forward, would be to add an additional linear constraint that `t >= 0`, instead of modifying the bounds, but I don't have the time to implement.

I this method tested/used by JuMP?

joehuchette:
Not by JuMP, I'm calling it directly in [Cerberus](https://github.com/joehuchette/Cerberus.jl). I'm not sure if this is being tested in MOI.Tests. Should I add a unit test in this repo?

joehuchette:
Screen Shot 2021-01-06 at 9 06 00 AM

odow:
A unit test would be good, if you have time.

blegat:
Maybe also add a comment explaining why Gurobi.jl defines this method (as you explain in the first comment of thi PR)

joehuchette:
Screen Shot 2021-01-07 at 9 22 23 AM

PR #392, opened 2021-02-04
[MOI] Allow accessing multiple results

odow:
Solves JuMP roadmap item "Accessing multiple results from solvers."

Here's what the test looks like at the JuMP level:
```Julia
using JuMP, Gurobi, Random, Test

function _is_binary(x; atol = 1e-6)
    return isapprox(x, 0; atol = atol) || isapprox(x, 1; atol = atol)
end

N = 30
Random.seed!(1)
item_weights, item_values = rand(N), rand(N)

model = Model(Gurobi.Optimizer)
@variable(model, x[1:N], Bin, start = 0)
@constraint(model, item_weights' * x <= 10)
@objective(model, Max, item_values' * x)
optimize!(model)
RC = result_count(model)
@test RC > 1

for n in [0, RC + 1]
    @test_throws MOI.ResultIndexBoundsError value.(x; result = n)
    @test_throws MOI.ResultIndexBoundsError objective_value(model; result = n)
end

for n = 1:RC
    xn = value.(x; result = n)
    @test all(_is_binary, xn)
    @test isapprox(
        objective_value(model; result = n),
        item_values' * xn,
        atol=1e-6,
    )
end
```

Tests pass:

image

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -2533,32 +2533,21 @@ function _has_primal_ray(model::Optimizer)
     return ret == 0
 end
 
-function _get_dbl_attr_variable(
-    model::Optimizer, key::String, x::MOI.VariableIndex
+function MOI.get(
+    model::Optimizer, attr::MOI.VariablePrimal, x::MOI.VariableIndex
 )
+    _throw_if_optimize_in_progress(model, attr)
+    MOI.check_result_index_bounds(model, attr)
+    key = model.has_unbounded_ray ? "UnbdRay" : "X"
+    if attr.N > 1
+        MOI.set(model, MOI.RawParameter("SolutionNumber"), attr.N - 1)
+        key = "Xn"
+    end
     col = Cint(column(model, x) - 1)
     valueP = Ref{Cdouble}()
     ret = GRBgetdblattrelement(model, key, col, valueP)
     _check_ret(model, ret)
     return valueP[]
-end
-
-function MOI.get(
-    model::Optimizer, attr::MOI.VariablePrimal, x::MOI.VariableIndex
-)
-    _throw_if_optimize_in_progress(model, attr)
-    if model.has_unbounded_ray
-        return _get_dbl_attr_variable(model, "UnbdRay", x)
-    end
-    result_count = MOI.get(model, MOI.ResultCount())
-    if !(1 <= attr.N <= result_count)
-        throw(MOI.ResultIndexBoundsError(attr, result_count))
-    elseif attr.N == 1
-        return _get_dbl_attr_variable(model, "X", x)
-    else
-        MOI.set(model, MOI.RawParameter("SolutionNumber"), attr.N - 1)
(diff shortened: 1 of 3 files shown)


src/MOI_wrapper.jl
```diff
@@ -2533,17 +2533,34 @@ function _has_primal_ray(model::Optimizer)
...
 end
 
+function MOI.get(
+    model::Optimizer, attr::MOI.VariablePrimal, x::MOI.VariableIndex
+)
+    _throw_if_optimize_in_progress(model, attr)
+    if model.has_unbounded_ray
+        return _get_dbl_attr_variable(model, "UnbdRay", x)
+    end
+    result_count = MOI.get(model, MOI.ResultCount())
```
blegat, line 2553: Uses `MOI.check_result_index_bounds`

src/MOI_wrapper.jl
```diff
@@ -2533,17 +2533,34 @@ function _has_primal_ray(model::Optimizer)
...
+    if !(1 <= attr.N <= result_count)
+        throw(MOI.ResultIndexBoundsError(attr, result_count))
+    elseif attr.N == 1
+        return _get_dbl_attr_variable(model, "X", x)
+    else
+        MOI.set(model, MOI.RawParameter("SolutionNumber"), attr.N - 1)
```
blegat, line 2559: Shouldn't we set this back to 1 afterwards ?

blegat: Ah no, I guess it only matters if you use `Xn`

src/MOI_wrapper.jl
```diff
@@ -2760,11 +2777,21 @@ end
 
 function MOI.get(model::Optimizer, attr::MOI.ObjectiveValue)
     _throw_if_optimize_in_progress(model, attr)
-    MOI.check_result_index_bounds(model, attr)
-    valueP = Ref{Cdouble}()
-    ret = GRBgetdblattr(model, "ObjVal", valueP)
-    _check_ret(model, ret)
-    return valueP[]
+    result_count = MOI.get(model, MOI.ResultCount())
```
blegat, line 2780: Uses MOI.check_result_index_bounds

src/MOI_wrapper.jl
```diff
@@ -2533,17 +2533,34 @@ function _has_primal_ray(model::Optimizer)
...
     _check_ret(model, ret)
     return valueP[]
 end
 
+function MOI.get(
+    model::Optimizer, attr::MOI.VariablePrimal, x::MOI.VariableIndex
+)
+    _throw_if_optimize_in_progress(model, attr)
+    if model.has_unbounded_ray
+        return _get_dbl_attr_variable(model, "UnbdRay", x)
```
blegat, line 2551: We should only do that if `attr.N` is 1 I suppose. and we should also do `MOI.check_result_index_bounds` before that

odow:
Oops. Yeah there was a much simpler implementation.

Issue #397, opened 2021-03-02
Add support for MEMORY_LIMIT termination status


henriquebecker91:
My situation: I am writing a paper that compares many formulations of the literature for a specific problem in the same computer setup (to be fair) and using both CPLEX and Gurobi (to be complete). Some formulations do not do well with some instances from literature datasets, and ask for more memory the computer setup has available. I do not want: to change the formulations (they are from the literature and should be executed as is); to buy more memory (the computer setup is from the university, and some instances will probably have problems even if I had the maximum allowed by the setup); to manually verify if a pair of formulation/instance has memory problems (they are thousands). Theoretically, I can solve this in CPLEX setting the memory limits for the branch-and-bound tree: this make it not throw any exceptions but instead return a different termination status. However, not every solver has memory limits or act the same way, in fact, Gurobi has no configurable memory limit, and instead will throw an exception when it perceive it will need more memory than the one provided by the system. Gurobi.jl wrapper actually deals with this kinda gracefully, capturing the exception and throwing it again. So I can theoretically capture the exception, and if it checks as out-of-memory I can mark this on the output and continue with the experiment. 

My questions: Can I depend on the exception thrown by a solver wrapper? (Can it be changed in a patch version? Or minor one after 1.0.0?) Will the solver wrappers maintained by jump-dev follow the same string pattern ("$SOLVER_NAME Error $ERROR_NUMBER: $ERROR_MESSAGE_FROM_SOLVER_IF_ANY") and be of type `ErrorException`? What is the rationale for having something as termination status or as exception?

Sorry if this is not the right channel, but I was not sure where to ask about the solver wrappers maintained by jump-dev in general (i.e., not about a specific solver wrapper).

henriquebecker91:
Okay. I will probably submit a draft of the PR tomorrow. I was thinking of doing the way you outlined but was worried it would be considered too hacky.

Do you think there may be other points, outside of termination status, that may need to check this flag to not trip over invalid state?

henriquebecker91:
The PR is #398.

odow:
Closed by #398

PR #398, opened 2021-03-03
Add fake termination status to Gurobi (so it become compliant with MOI.MEMORY_LIMIT)

henriquebecker91:
This PR intends to solve #397. The PR creates a generic mechanism that allows Gurobi Error Codes  to interpreted as termination status by the Gurobi wrapper. For now, it is used only for transforming `OUT_OF_MEMORY` in `MOI.MEMORY_LIMIT`, but new conversions only need to add a new element to a global internal Dict. The mechanism consist in one extra field for `Gurobi.Optimizer` (called `fake_termination_status`), a new internal function `_check_ret_GRBoptimize`, a new internal global `_FAKE_TERMINATION_STATUS`, and changes to `_raw_status`, `Optimizer` (constructor), `is_empty`, `empty!`, and `optimize!`.

The code is well commented. I did not add tests. If tests are deemed necessary, then I would like that someone more familiar with exception testing inside `Gurobi.jl` gave me some pointers.

The current code give some errors:

[](https://postimg.cc/DSC05WHQ)

but they are the same obtained in the master branch:

[](https://postimg.cc/PPpSdZzJ)

and probably should be fixed in other PR:

[](https://postimg.cc/wyNB6wvF)

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -302,8 +302,9 @@ end
 # If you add a new error code that, when returned by GRBoptimize,
 # should be treated as a TerminationStatus by MOI, to the global
 # below, then the rest of the code should pick up on this seamlessly.
-const _FAKE_TERMINATION_STATUS = Dict{Cint, Tuple{Any, String}}([
-    Cint(10001) => (MOI.MemoryLimit, "Available memory was exhausted."),
+const _FAKE_TERMINATION_STATUS =
+Dict{Cint, Tuple{MOI.TerminationStatusCode, String}}([
+    Cint(10001) => (MOI.MEMORY_LIMIT, "Available memory was exhausted."),
 ])
 
 # Same as _check_ret, but deals with the `fake_termination_status` machinery.
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -187,6 +187,23 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     name_to_constraint_index::Union{Nothing,
         Dict{String, Union{Nothing, MOI.ConstraintIndex}}}
 
+
+    # Gurobi does not have a configurable memory limit (different of time),
+    # but it does detect when it needs more memory than it is available,
+    # and it stops the optimization returning a specific error code.
+    # This is a different mechanism than Gurobi "Status" (that is used for
+    # reporting why an optimization finished) and, in fact, may be triggered in
+    # other cases than optimization (for example, when assembling the model).
+    # For convenience, and homogeinity with other solvers, when a call to
+    # `GRBoptimize` return an error that would be considered a
+    # `TerminationStatus` by MOI, we do not thrown an exception (like we do
+    # for errors) but instead we set the field below `fake_termination_status`
+    # to the error code. Any non-zero value in this field overrides the query
+    # to Gurobi "Status" and instead we return the appropriate
+    # `MOI.TerminationStatus`. Just before calling `GRBoptimize` we change
+    # this to zero again.
+    fake_termination_status::Cint
+
     # These two flags allow us to distinguish between FEASIBLE_POINT and
     # INFEASIBILITY_CERTIFICATE when querying VariablePrimal and ConstraintDual.
     has_unbounded_ray::Bool
@@ -248,6 +265,9 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
         model.quadratic_constraint_info = Dict{Int, _ConstraintInfo}()
         model.sos_constraint_info = Dict{Int, _ConstraintInfo}()
         model.indicator_constraint_info = Dict{Int, _ConstraintInfo}()
+
+        model.fake_termination_status = Cint(0)
+
         model.callback_variable_primal = Float64[]
         MOI.empty!(model)
         finalizer(model) do m
@@ -275,6 +295,23 @@ function _check_ret(model::Optimizer, ret::Cint)
     if ret != 0
         msg = unsafe_string(GRBgetmerrormsg(model))
         throw(ErrorException("Gurobi Error $(ret): $(msg)"))
+    end
+    return
+end
+
+# If you add a new error code that, when returned by GRBoptimize,
+# should be treated as a TerminationStatus by MOI, to the global
+# below, then the rest of the code should pick up on this seamlessly.
+const _FAKE_TERMINATION_STATUS = Dict{Cint, Tuple{Any, String}}([
+    Cint(10001) => (MOI.MemoryLimit, "Available memory was exhausted."),
+])
+
+# Same as _check_ret, but deals with the `fake_termination_status` machinery.
+function _check_ret_GRBoptimize!(model, ret::Cint)
+    if haskey(_FAKE_TERMINATION_STATUS, ret)
+        model.fake_termination_status = ret
+    else
+        _check_ret(model, ret)
     end
     return
 end
@@ -346,6 +383,7 @@ function MOI.empty!(model::Optimizer)
     empty!(model.indicator_constraint_info)
     model.name_to_variable = nothing
     model.name_to_constraint_index = nothing
+    model.fake_termination_status = Cint(0)
     model.has_unbounded_ray = false
     model.has_infeasibility_cert = false
     empty!(model.callback_variable_primal)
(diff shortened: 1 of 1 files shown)


src/MOI_wrapper.jl
```diff
@@ -187,6 +187,23 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
...
+    # other cases than optimization (for example, when assembling the model).
+    # For convenience, and homogeinity with other solvers, when a call to
+    # `GRBoptimize` return an error that would be considered a
+    # `TerminationStatus` by MOI, we do not thrown an exception (like we do
+    # for errors) but instead we set the field below `fake_termination_status`
+    # to the error code. Any non-zero value in this field overrides the query
+    # to Gurobi "Status" and instead we return the appropriate
+    # `MOI.TerminationStatus`. Just before calling `GRBoptimize` we change
+    # this to zero again.
+    fake_termination_status::Cint
```
odow, line 205: `ret_GRBoptimize::Cint`?

src/MOI_wrapper.jl
```diff
@@ -248,6 +265,9 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
         model.quadratic_constraint_info = Dict{Int, _ConstraintInfo}()
         model.sos_constraint_info = Dict{Int, _ConstraintInfo}()
         model.indicator_constraint_info = Dict{Int, _ConstraintInfo}()
+
+        model.fake_termination_status = Cint(0)
```
odow, line 269: No need to set here. It gets set in `MOI.empty!`.

src/MOI_wrapper.jl
```diff
@@ -279,6 +299,24 @@ function _check_ret(model::Optimizer, ret::Cint)
     return
 end
 
+# If you add a new error code that, when returned by GRBoptimize,
+# should be treated as a TerminationStatus by MOI, to the global
+# below, then the rest of the code should pick up on this seamlessly.
+const _FAKE_TERMINATION_STATUS =
+Dict{Cint, Tuple{MOI.TerminationStatusCode, String}}([
+    Cint(10001) => (MOI.MEMORY_LIMIT, "Available memory was exhausted."),
+])
```
odow, line 308:
```Julia
const _ERROR_CODES = Dict{Cint,Tuple{MOI.TerminationStatusCode,String}}(
    Cint(10001) => (MOI.MEMORY_LIMIT, "Available memory was exhausted."),
)
```

src/MOI_wrapper.jl
```diff
@@ -401,6 +441,11 @@ function _update_if_necessary(model::Optimizer)
     if model.needs_update
         sort!(model.columns_deleted_since_last_update)
         for var_info in values(model.variable_info)
+            # The trick here is: searchsortedlast returns, in O(log n), the
+            # last index with a column smaller than var_info.column, over
+            # columns_deleted_since_last_update this is the same as the number
+            # of columns deleted before it, and how much its value need to be
+            # shifted.
```
odow, line 448: Please add in a different PR. It's unrelated.

src/MOI_wrapper.jl
```diff
@@ -2403,8 +2448,13 @@ function MOI.optimize!(model::Optimizer)
     #
     # TODO(odow): Julia 1.5 exposes `Base.exit_on_sigint(::Bool)`.
     ccall(:jl_exit_on_sigint, Cvoid, (Cint,), false)
+    # Before calling GRBoptimize we clean the model from any fake
+    # termination status that exists for the MOI wrapper but it is
+    # not really supported by Gurobi. `_check_ret_GRBoptimize!`
+    # will set it again if it is the case.
+    model.fake_termination_status = Cint(0)
     ret = GRBoptimize(model)
-    _check_ret(model, ret)
```
odow, line 2407:
```julia
model.retGRBoptimize = GRBoptimize(model)
_check_ret_GRBoptimize(model)
```

src/MOI_wrapper.jl
```diff
@@ -279,6 +299,24 @@ function _check_ret(model::Optimizer, ret::Cint)
...
+# If you add a new error code that, when returned by GRBoptimize,
+# should be treated as a TerminationStatus by MOI, to the global
+# below, then the rest of the code should pick up on this seamlessly.
+const _FAKE_TERMINATION_STATUS =
+Dict{Cint, Tuple{MOI.TerminationStatusCode, String}}([
+    Cint(10001) => (MOI.MEMORY_LIMIT, "Available memory was exhausted."),
+])
+
+# Same as _check_ret, but deals with the `fake_termination_status` machinery.
+function _check_ret_GRBoptimize!(model, ret::Cint)
```
odow, line 311:
```Julia
function _check_ret_GRBoptimize(model::Optimizer)
    if !haskey(_ERROR_CODES, model.ret_GRBoptimize)
        _check_ret(model, model.ret_GRBoptimize)
    end
    return
end
```

src/MOI_wrapper.jl
```diff
@@ -2448,6 +2498,9 @@ const _RAW_STATUS_STRINGS = [
 ]
 
 function _raw_status(model::Optimizer)
+    if !iszero(model.fake_termination_status)
```
odow, line 2501:
```Julia
if haskey(_ERROR_CODES, model.ret_GRBoptimize)
    return _ERROR_CODES[model.ret_GRBoptimize]
end
```

odow (commented):
Nice. Are there any other return codes we should catch?

odow:
Are you on a new release of Gurobi? Good to see they fixed it.

henriquebecker91:
> Are you on a new release of Gurobi? Good to see they fixed it.

Sorry, fixed what exactly? The Farkas dual? I am using Gurobi 9.1.1.

odow:
There was a bug in Gurobi. But they fixed it which is why you now see "Got correct result"

henriquebecker91:
About the changes, I have no problem with the purely stylistic ones, and will probably make a new commit soon to solve them. However, some of the changes of names and how conditionals are done are pointing could be looked at with a little more attention.

`fake_termination_status` -> `ret_GRBoptimize::Cint`
```
model.retGRBoptimize = GRBoptimize(model)
_check_ret_GRBoptimize(model)
```
`_FAKE_TERMINATION_STATUS` -> `_ERROR_CODES`

Its seems to me that, instead of making this field only for the error codes that may be interpreted as `TerminationStatus`es, the idea is always saving the error code of the GRBoptimize, and then, if the termination status is queried, we look at the `Dict`. This alone is ok, even more elegant than what I have initially thought (I mean, probably `haskey` may end up called multiple times but then the overhead is minimal). However, I think in this case the `Dict` really should not be called `_ERROR_CODES` that gives the impression that all Gurobi `_ERROR_CODES` are stored there. Maybe `_ERROR_TO_STATUS` instead? It is shorter and not so misleading.

odow:
> Maybe _ERROR_TO_STATUS instead?

Sure.

henriquebecker91:
Made the changes, the tests continue to pass:

[](https://postimg.cc/1fp19fQG)

src/MOI_wrapper.jl
```diff
@@ -279,6 +294,22 @@ function _check_ret(model::Optimizer, ret::Cint)
     return
 end
 
+# If you add a new error code that, when returned by GRBoptimize,
+# should be treated as a TerminationStatus by MOI, to the global `Dict`
+# below, then the rest of the code should pick up on this seamlessly.
+const _ERROR_TO_STATUS = Dict{Cint, Tuple{MOI.TerminationStatusCode, String}}([
+    # TerminationStatus, RawStatusString
+    Cint(10001) => (MOI.MEMORY_LIMIT, "Available memory was exhausted."),
```
odow, line 302: Can you use instead `GRB_ERROR_OUT_OF_MEMORY => `: https://github.com/jump-dev/Gurobi.jl/blob/a995c081b7717b2f1a3e842d4b9ca51fa817d144/src/gen91/libgrb_common.jl#L11

odow (commented):
Looks good.

henriquebecker91:
Made the change, tests continue the same:

[](https://postimg.cc/sMCQXnNM)

src/MOI_wrapper.jl
henriquebecker91: Done.

odow (approved):

PR #403, opened 2021-04-04
Allow enable_interrupts=false

odow:
Motivated by this thread: https://discourse.julialang.org/t/jump-optimize-call-will-very-rarely-violate-models-time-limit-parameters-by-huge-factors/58061/17

cc @Antelling


odow:
image

PR #409, opened 2021-05-12
Explicit tests for MOI.TerminationStatus with infeasible statuses

dourouc05:
This ensures that an issue like https://github.com/jump-dev/CPLEX.jl/pull/366/files doesn't sneak in.

Another solution may be to move some IIS tests to MOIT, but I don't know how to test `Gurobi.ConflictStatus` (or the CPLEX equivalent) in that case.

### DIFF
--- a/test/MOI/MOI_wrapper.jl
+++ b/test/MOI/MOI_wrapper.jl
@@ -302,6 +302,7 @@ function test_Conflict_refiner_bound_bound()
 
     # Once it's called, no problem.
     MOI.compute_conflict!(model)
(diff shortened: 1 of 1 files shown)


odow:
Did you mean to open the two PRs? I think we just need this one, correct?

dourouc05:
Actually, I meant to make two PRs with different content, but I mixed the branches locally… The other PR has minor fixes that this one doesn't have, I'll try to fix the mixup?

dourouc05:
The PRs should now be clearly separate. Here is a screenshot of the tests passing: 

image

odow (approved):

PR #427, opened 2021-11-01
Run the lazy constraints callback at MIPNODE

### LINKED ISSUE #426: MOI lazy constraints callback not called at MIPNODE
From the [JuMP solver independent callbacks examples](https://jump.dev/JuMP.jl/stable/manual/callbacks/#Things-you-can-and-cannot-do-during-solver-independent-callbacks) it should be possible for fractional solutions to reach a lazy constraints callback.

MOI_callbacks.jl only calls the lazy constraints callback (if it exists) when where=MIPSOL, so fractional node solutions can never reach the callback. This is contrary to the behaviour of the example code when using GLPK.

https://github.com/jump-dev/Gurobi.jl/blob/374a569f71dffa178092d73fda58f93a537ae425/src/MOI_wrapper/MOI_callbacks.jl#L126-L152

```
function my_callback_function(cb_data)
    status = callback_node_status(cb_data, model)
    if status == MOI.CALLBACK_NODE_STATUS_FRACTIONAL
        # `callback_value(cb_data, x)` is not integer (to some tolerance).
        # <-- NOT reachable with Gurobi.jl backend -->
    elseif status == MOI.CALLBACK_NODE_STATUS_INTEGER
        # `callback_value(cb_data, x)` is integer (to some tolerance).
    else
        @assert status == MOI.CALLBACK_NODE_STATUS_UNKNOWN
        # `callback_value(cb_data, x)` might be fractional or integer.
    end
    ...
end
MOI.set(model, MOI.LazyConstraintCallback(), my_callback_function)
```

simonbowly:
Executes the lazy constraints callback (if defined) when where=MIPNODE so that the callback is executed at fractional values. Added a test using callback_knapsack_model to ensure some fractional solutions reach the callback in this example.

Note: branched this from v0.9.14 as v0.10.0 forces a downgrade to JuMP 0.18

### DIFF
--- a/src/MOI_wrapper/MOI_callbacks.jl
+++ b/src/MOI_wrapper/MOI_callbacks.jl
@@ -146,6 +146,10 @@ function _default_moi_callback(model::Optimizer)
                 model.callback_state = _CB_HEURISTIC
                 model.heuristic_callback(cb_data)
             end
+            if model.lazy_callback !== nothing
+                model.callback_state = _CB_LAZY
+                model.lazy_callback(cb_data)
+            end
         end
         model.callback_state = _CB_NONE
     end
--- a/test/MOI/MOI_callbacks.jl
+++ b/test/MOI/MOI_callbacks.jl
@@ -106,6 +106,32 @@ function test_lazy_constraint_callback()
     @test MOI.get(model, MOI.VariablePrimal(), y) == 2
 end
 
+function test_lazy_constraint_callback_fractional()
+    # callback_simple_model() is not large enough to see MIPNODE callbacks.
+    model, x, item_weights = callback_knapsack_model()
+    lazy_called_integer = false
+    lazy_called_fractional = false
+    MOI.set(
+        model,
+        MOI.LazyConstraintCallback(),
+        cb_data -> begin
+            status = MOI.get(
+                model,
+                MOI.CallbackNodeStatus(cb_data),
+            )::MOI.CallbackNodeStatusCode
+            if status == MOI.CALLBACK_NODE_STATUS_INTEGER
+                lazy_called_integer = true
+            elseif status == MOI.CALLBACK_NODE_STATUS_FRACTIONAL
+                lazy_called_fractional = true
+            end
+        end,
+    )
+    @test MOI.supports(model, MOI.LazyConstraintCallback())
+    MOI.optimize!(model)
+    @test lazy_called_integer
+    @test lazy_called_fractional
+end
+
 function test_lazy_constraint_callback_OptimizeInProgress()
     model, x, y = callback_simple_model()
     MOI.set(


odow:
Hi @simonbowly,

Yes, I think I originally only called it at MIPSOL because we didn't have a `CallbackNodeStatus` attribute at first.

The JuMP documentation does suggest that it may be called at fractional nodes:
image

Re the 0.10.0 stuff, yes, we're very close (https://github.com/jump-dev/JuMP.jl/milestone/12) to releasing a new version of JuMP. Gurobi 0.10 works with MathOptInterface 0.10, but JuMP 0.21 relies on MathOptInterface 0.9.

Can you post a screenshot of the tests passing? We can't run Gurobi on CI due to licensing issues.

odow:
p.s. if you want to chat about JuMP + Gurobi.jl, we're very keen to have some more official support from Gurobi on this. The licensing issue is a major pain point.

simonbowly:
Sure thing - test output below for this branch. I see, without CallbackNodeStatus it makes sense that you would want to avoid this call at MIPNODE. And I guess this could break code not already doing this status check.

image

simonbowly:
This came up in the context of my academic work by the way.  But I hadn't realised licensing was blocking CI for you - I'll email you directly about this as I'm sure we can sort something out!

odow:
@blegat do you regard this as breaking? I guess if we push this out as 0.10.1, must users will be updating from JuMP 0.21  and Gurobi 0.9 to JuMP 0.22 and Gurobi 0.10.1 so it's a good time to do this.

odow:
> This came up in the context of my academic work by the way

The other option, and one that we should encourage, is for users to write solver-specific callbacks whenever the MOI defaults aren't suitable:

https://github.com/jump-dev/Gurobi.jl#callbacks

odow:
So I knew I'd gone over this at some point. My original implementation had it, but I removed it here:
https://github.com/jump-dev/Gurobi.jl/pull/346/commits/8091d26926700461c9106c683bbcd728259254f2#diff-ee94cc576005eb58078f7a3fc01b0f63a66806f019d6b4513ca8f7062f2dd4b8L120-R128

I don't remember the reasoning, but we do the same in CPLEX:
https://github.com/jump-dev/CPLEX.jl/blob/b823bbb90e55046220925c6ba8895b195669f270/src/MOI/MOI_callbacks.jl#L194-L204

What's the motivation for adding lazy constraints to fractional solutions?

simonbowly:
I see - this is for cases where knowledge of fractional solutions might allow the user to identify a necessary lazy constraint earlier in the solve, before an integer solution which violates it arises.  In some sense it's equivalent to a user cut, but the behaviour can be slightly different.

If only calling the lazy constraint callback at integer solutions is the MOI default, this is also fine, but the JuMP docs for solver-independent lazy constraint callbacks (with callback_node_status) imply that fractional node relaxations should reach this callback. This does occur when using GLPK. I expected this to be consistent across solvers and assumed Gurobi needed to be brought into line with the others, sorry!

simonbowly:
Now I see that GLPK doesn't distinguish these cases in its own callback reason codes, and GLPK.jl has to check integrality to create the callback status, which would explain the difference.

odow:
The operative word in the JuMP docs is "may". It's painfully difficult to ensure consistency across the solvers that implement callbacks.

There's essentially no guarantee that JuMP's solver-independent callbacks behave the same across solvers. All it is is a nice syntax for some common callbacks; the actual behavior is solver-dependent.

That said, I don't see a good reason not to call the lazy constraint at fractional nodes, especially since it's documented and we now have the status mechanism. I'll wait for @blegat to chime in, but this looks good to me.

odow (approved):

blegat:
> @blegat do you regard this as breaking? I guess if we push this out as 0.10.1, must users will be updating from JuMP 0.21 and Gurobi 0.9 to JuMP 0.22 and Gurobi 0.10.1 so it's a good time to do this.

Indeed, we could also simply release v0.11.

> That said, I don't see a good reason not to call the lazy constraint at fractional nodes, especially since it's documented and we now have the status mechanism.

I agree, it's also good to remove this difference with GLPK.

blegat (approved):

odow:
Thanks @simonbowly. My email is o.dowson@gmail.com if you want to discuss the licensing issue.

PR #428, opened 2021-11-04
Add continuous integration

odow:
No idea if this will work, but let's see.

cc @simonbowly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: CI
+on:
+  push:
+    branches: [master]
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  test:
+    name: 'Gurobi'
+    runs-on: 'ubuntu-latest'
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: '1'
+          arch: 'x64'
+      - uses: actions/cache@v1
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: julia-actions/julia-runtest@v1
+      - uses: julia-actions/julia-processcoverage@v1
+      - uses: codecov/codecov-action@v1
+        with:
+          file: lcov.info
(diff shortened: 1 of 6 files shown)


odow:
Success 
image

odow:
Artifacts are pretty cool: https://pkgdocs.julialang.org/v1/artifacts/. They make including downloads like this trivial!

simonbowly:
Nice one!  I made some progress testing the license - evidently I didn't quite configure it right, but copying in the license file from a repo secret mostly does the trick.

image

odow:
@simonbowly I've invited you as an admin user. That should mean you have permission to upload secrets. Let me know if that isn't the case.

simonbowly:
Cheers - I'm just testing using my own license at the moment so will wait for yours to be issued (almost sorted I think).

simonbowly:
Got the license configuration worked out - tests running ok on my master branch.  Now just need to sort out your license (getting there sorry!)  I had to change the artifacts usage slightly to get this running, not really sure why...

image

odow:
> I had to change the artifacts usage

You probably have an existing `deps.jl` file in your local clone of the repository?

You should have permission to push new commits to this branch, or edit the files via the web editor.

simonbowly:
Done - I pushed just the CI change to this PR and now see the Artifacts error at build time. "Artifacts not defined"?

odow:
Oops. I was being too clever when I squashed all my test commits. Working now.

Did you upload the secret? I can't see it

image

simonbowly:
Nice one. I didn't yet - I was using my own license key to test on my fork. Would rather wait for the one issued specifically for JuMP if that's ok. But this should otherwise work once that is sorted.

odow:
@simonbowly looks like I'll need your input on this:

image

Did you test a WSL on your fork? Or some other type of license?

odow:
@simonbowly and I discussed the license issue offline. It's now working.

However: @simonbowly when are the licenses validated? I'm seeing a lot of:
image
I assumed it would be whenever a new `Env` is created, but it seems that it might be on every call to optimize? Or periodically?

odow:
It only took 8.5 years, but we now have CI! That coverage is a little shoddy though.

simonbowly:
Great!  Thanks for all your work on this Oscar.

Re: the license checks, for WLS the license token needs to be refreshed periodically from the server.  The checks might occur periodically as well as on various API calls including .update() and .optimize() so I guess in tests it occurs pretty often.  It is probably not communicating with the server for every one of those messages though.

szcf-weiya:
Thanks for the great package! I am using Gurobi.jl in my project, and also want to conduct a continuous integration in github actions. At first, it throws `No license` error,
image

then I tried to use a WSL license like what you did here, but now it throws
image

it seems that you also encountered the same issue (https://github.com/jump-dev/Gurobi.jl/pull/428#issuecomment-963486321); how did you finally manage it? Anything I have missed?

odow:
> how did you finally manage it?

We use an undocumented/still-in-development feature, so for now there is probably no way for you to use Gurobi.jl in GitHub actions. If you have a commercial license, you should contact your support rep and they may be able to help you.

szcf-weiya:
Got it, thanks for your explanation. Also just noticed README.md wrote
> However, we thank Gurobi for providing us with a license to test Gurobi.jl on GitHub.

PR #434, opened 2021-11-12
Fix failed MOI tests

metab0t:
I applied the same tricks used in Xpress.jl here to fix 3 failed MOI tests.
The `test_solve_conflict_feasible` fails because Gurobi cannot compute IIS on a feasible model

### DIFF
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -3423,9 +3423,7 @@ function MOI.get(model::Optimizer, ::MOI.ListOfConstraintTypesPresent)
 end
 
 function MOI.get(model::Optimizer, ::MOI.ObjectiveFunctionType)
-    if !model.is_objective_set
-        return nothing
-    elseif model.objective_type == _SINGLE_VARIABLE
+    if model.objective_type == _SINGLE_VARIABLE
         return MOI.VariableIndex
     elseif model.objective_type == _SCALAR_AFFINE
         return MOI.ScalarAffineFunction{Float64}
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -679,10 +679,16 @@ function MOI.get(model::Optimizer, ::MOI.ListOfModelAttributesSet)
     return attributes
 end
 
-function MOI.get(model::Optimizer, ::MOI.ListOfConstraintAttributesSet{F, S}) where {S, F}
+function MOI.get(
+    model::Optimizer,
+    ::MOI.ListOfConstraintAttributesSet{F,S},
+) where {S,F}
     ret = MOI.AbstractConstraintAttribute[]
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -3409,7 +3417,7 @@ function MOI.get(model::Optimizer, ::MOI.ListOfConstraintTypesPresent)
 end
 
 function MOI.get(model::Optimizer, ::MOI.ObjectiveFunctionType)
-    if model.is_feasibility
+    if !model.is_objective_set
```
odow, line 3420: I think in other wrappers I removed the `return nothing`. If the objective isn't set, getting `MOI.ObjectiveFunctionType` doesn't have a valid return, so can return anything.

odow:
Can you fix the format-check please?

src/MOI_wrapper/MOI_wrapper.jl
metab0t: Done

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -661,7 +662,13 @@ function MOI.get(model::Optimizer, ::MOI.ListOfVariableAttributesSet)
...
-    attributes = Any[MOI.ObjectiveSense()]
+    if MOI.is_empty(model)
+        return Any[]
+    end
+    attributes = Any[]
+    if model.objective_sense !== nothing
+        push!(attributes, MOI.ObjectiveSense())
+    end
     typ = MOI.get(model, MOI.ObjectiveFunctionType())
     if typ !== nothing
```
odow, line 673: Is this passing tests? It should be
```Julia
if model.is_objective_set
    F = MOI.get(model, MOI.ObjectiveFunctionType())
    push!(attributes, MOI.ObjectiveFunction{F}())
end
```
(ObjectiveFunctionType now never returns `nothing`.)

metab0t: Yes, tests passed even if this is not corrected. Maybe a new MOI test should be added?

metab0t:
I think `objective_type` should be a nullable type like `objective_sense`. 

`is_objective_set` can be discarded and `objective_type = nothing` represents no objective function is set.

The complicated thing is here. We need to check the corresponding variable of changed coefficient and set `objective_type` if `objective_type = _SINGLE_VARIABLE` currently.

```julia
function MOI.modify(
    model::Optimizer,
    ::MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}},
    chg::MOI.ScalarCoefficientChange{Float64},
)
    ret = GRBsetdblattrelement(
        model,
        "Obj",
        Cint(column(model, chg.variable) - 1),
        chg.new_coefficient,
    )
    _check_ret(model, ret)
    model.is_objective_set = true
    _require_update(model)
    return
end
```

odow:
I think this is okay for now.

PR #439, opened 2021-11-22
Always write deps.jl so Gurobi.jl is recompiled on Pkg.build

### LINKED ISSUE #438: Force recompilation when deps.jl changes
It's really nice that this is now downloading an artifact, however it's really screwing up a case in which I have an older computer server running which is incompatible with 0.9.5.

I was hoping that by setting `GUROBI_HOME` when Gurobi.jl gets built it would use that and not try to use the artifact, however I couldn't seem to get it to do this.

I suggest
- If no `GUROBI_HOME` is set, use artifacts as it currently does.
- If `GUROBI_HOME` is set, skip artifacts and just use that.

odow:
Closes #438

@ExpandingMan can you test this branch please?

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -161,7 +161,7 @@ if haskey(ENV, "GUROBI_JL_SKIP_LIB_CHECK")
     # Skip!
 elseif get(ENV, "JULIA_REGISTRYCI_AUTOMERGE", "false") == "true"
     write_depsfile("julia_registryci_automerge")
-elseif !found &&Sys.islinux()
+elseif !found && Sys.islinux()
     open(DEPS_FILE, "w") do io
         println(io, "# No libgurobi constant; we're using the Artifact.")
     end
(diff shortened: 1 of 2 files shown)


ExpandingMan:
I have off this week, so I won't get a chance to check this on a real gurobi environment until next week.  Feel free to merge and I'll check it out next week.  If you prefer to wait, I'll let you know what I find next week.  Either way, I'll get back to you next week.

Thanks for taking the time to fix this the right way!

odow:
I'll hold off merging because I haven't tested this on a linux system (I only have Mac).

odow:
Any chance someone with Gurobi installed on a linux machine could test this?

ExpandingMan:
I apologize for disappearing on you, of course I forgot to come back and check this, and for whatever reason failed to notice the github notification on your first reminder.

Anyway, this seems to be working as intended!  First I deleted the gurobi artifact, and ran `Pkg.build("Gurobi")`.  The artifact re-appeared, but I deleted it to make sure it was using the set `GUROBI_HOME` and it still worked.  So, I'm not 100% sure this solves whatever the root problem was that prompted me to post that issue, but it certainly seems to work with either the artifact or `GUROBI_HOME`.

Thanks agian.

odow:
> I apologize for disappearing on you

You're a volunteer, so no need to apologize. Thanks for checking.

PR #448, opened 2022-02-13
fix error due to adding null callback

### LINKED ISSUE #395: Undesired callback behavior
Here's a MWE of the issue. I hope it is pretty straightforward. 
```julia
using JuMP
using Gurobi
m = Model(Gurobi.Optimizer)
@variable(m, x >= 1)
@objective(m, Min, x)
optimize!(m)
```
Running just the above code, if I check ``` m.moi_backend.optimizer.model.generic_callback```, I see that there is already a callback there, even though I didn't add it myself:
``` julia
Gurobi._CallbackUserData(    sense  : minimize
    number of variables             = 1
    number of linear constraints    = 0
    number of quadratic constraints = 0
    number of sos constraints       = 0
    number of non-zero coeffs       = 0
    number of non-zero qp objective terms  = 0
    number of non-zero qp constraint terms = 0...
```
For some reason, running ```optimize!``` seems to just add a generic callback. 
This is problematic, because if I try to add a callback as following:
```julia
function cb_fn(cb_data)
    val = JuMP.callback_value(x)
    con = @build_constraint(x >= 2)
    MOI.submit(m, MOI.LazyConstraint(cb_data), con)
end
MOI.set(m, MOI.LazyConstraintCallback(), cb_fn)
optimize!(m)
```
It throws the following error, since it thinks that I have both a Gurobi.CallbackFunction and an MOI one:
``` julia
ERROR: Cannot use Gurobi.CallbackFunction as well as MOI.AbstractCallbackFunction
Stacktrace:
 [1] error(::String) at .\error.jl:33
 [2] _check_moi_callback_validity at C:\Users\berko\.julia\packages\Gurobi\Lci9Q\src\MOI_wrapper.jl:2372 [inlined]
 [3] optimize!(::Gurobi.Optimizer) at C:\Users\berko\.julia\packages\Gurobi\Lci9Q\src\MOI_wrapper.jl:2386
 [4] optimize!(::MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}) at C:\Users\iai\builds\InterpretableAI\SystemImage\SysImgBuilder\.julia\packages\MathOptInterface\ZJFKw\src\Bridges\bridge_optimizer.jl:264
 [5] optimize!(::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}) at C:\Users\iai\builds\InterpretableAI\SystemImage\SysImgBuilder\.julia\packages\MathOptInterface\ZJFKw\src\Utilities\cachingoptimizer.jl:215
 [6] optimize!(::Model, ::Nothing; bridge_constraints::Bool, ignore_optimize_hook::Bool, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at C:\Users\iai\builds\InterpretableAI\SystemImage\SysImgBuilder\.julia\packages\JuMP\qhoVb\src\optimizer_interface.jl:130
 [7] optimize! at C:\Users\iai\builds\InterpretableAI\SystemImage\SysImgBuilder\.julia\packages\JuMP\qhoVb\src\optimizer_interface.jl:106 [inlined] (repeats 2 times)
 [8] top-level scope at REPL[61]:1
```
Furthermore, if I try to add the callback without the first ``` optimize!``` to resolve the issue, it doesn't give the right solution, and returns 1 instead of 2! Please help. I'm using Gurobi 0.9.7 and JuMP 0.21.5.

chriscoey:
replaces #447 (CI would not run on PR from fork)
fixes #395

### DIFF
--- a/test/MOI/MOI_callbacks.jl
+++ b/test/MOI/MOI_callbacks.jl
@@ -83,10 +83,8 @@ function test_lazy_constraint_callback()
                 MOI.CallbackNodeStatus(cb_data),
             )::MOI.CallbackNodeStatusCode
             int_sol = round.(Int, [x_val, y_val])
-            if isapprox(int_sol, [x_val, y_val], atol = 1e-6)
-                @test status == MOI.CALLBACK_NODE_STATUS_INTEGER
-            else
-                @test status == MOI.CALLBACK_NODE_STATUS_FRACTIONAL
+            if status == MOI.CALLBACK_NODE_STATUS_INTEGER
+                @test isapprox(int_sol, [x_val, y_val], atol = 1e-6)
             end
             @test MOI.supports(model, MOI.LazyConstraint(cb_data))
             if y_val - x_val > 1 + 1e-6
@@ -299,10 +297,8 @@ function test_heuristic_callback()
                 MOI.CallbackNodeStatus(cb_data),
             )::MOI.CallbackNodeStatusCode
             int_sol = round.(Int, x_vals)
-            if isapprox(int_sol, x_vals, atol = 1e-6)
-                @test status == MOI.CALLBACK_NODE_STATUS_INTEGER
-            else
-                @test status == MOI.CALLBACK_NODE_STATUS_FRACTIONAL
+            if status == MOI.CALLBACK_NODE_STATUS_INTEGER
+                @test isapprox(int_sol, x_vals, atol = 1e-6)
             end
             if MOI.submit(
                 model,
--- a/test/MOI/MOI_callbacks.jl
+++ b/test/MOI/MOI_callbacks.jl
@@ -82,9 +82,8 @@ function test_lazy_constraint_callback()
                 model,
                 MOI.CallbackNodeStatus(cb_data),
             )::MOI.CallbackNodeStatusCode
-            if round.(Int, [x_val, y_val]) ≈ [x_val, y_val]
-                atol = 1e-6
-                @test status == MOI.CALLBACK_NODE_STATUS_INTEGER
+            if status == MOI.CALLBACK_NODE_STATUS_INTEGER
+                @test isapprox(round.(Int, [x_val, y_val]), [x_val, y_val], atol = 1e-5)
             else
                 @test status == MOI.CALLBACK_NODE_STATUS_FRACTIONAL
             end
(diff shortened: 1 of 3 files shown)


odow:
@simonbowly any chance you could take a look at our license?

image

I tried generating a new API key, but it's still tied to the expired license.

simonbowly:
@odow sorry about that, I just updated the license. Should be valid now but you might need to generate the API key again.

I'll follow up to see why the duration is so short.

odow:
That seems to have fixed it. 

> I'll follow up to see why the duration is so short.

Yes. Having a 2 week expiry isn't ideal...

odow:
@chriscoey I excluded some new tests from MOI about infeasibility certificates which you can ignore. 

But the heuristic callback test failing seems related?

chriscoey:
the heuristic callback test failing seems unrelated to this PR, but I pushed a fix for it here

test/MOI/MOI_callbacks.jl
```diff
@@ -82,9 +82,8 @@ function test_lazy_constraint_callback()
                 model,
                 MOI.CallbackNodeStatus(cb_data),
             )::MOI.CallbackNodeStatusCode
-            if round.(Int, [x_val, y_val]) ≈ [x_val, y_val]
-                atol = 1e-6
-                @test status == MOI.CALLBACK_NODE_STATUS_INTEGER
+            if status == MOI.CALLBACK_NODE_STATUS_INTEGER
+                @test isapprox(round.(Int, [x_val, y_val]), [x_val, y_val], atol = 1e-5)
```
odow, line 86: Ah. This looks like a formatting error from JuliaFormatter where the `atol` got dropped onto the wrong line.

The test should be
```Julia
               int_sol = round.(Int, [x_val, y_val])
               if isapprox(int_sol, [x_val, y_val], atol = 1e-6)
                 @test status == MOI.CALLBACK_NODE_STATUS_INTEGER
```

This revision isn't correct, because it's testing the reverse implication of what we want to test.

odow: I'll sort the edit

chriscoey: I changed the test because it seems like it would be numerically sensitive

chriscoey: it failed again. is there a reason it is testing the implication that way and not the other way?

odow: We want to check that if you are at an integer solution, Gurobi will report that the `CallbackNodeStatus` is `INTEGER`. The reverse implication is that if the node status is integer, then the solution is integral. But we might never hit the `INTEGER` status in our tests. 

This seems weird that we have an integer solution, but Gurobi is claiming it is at a MIPNODE not a MIPSOL. Perhaps I need to tighten the tolerances.

It's also weird that this was passing tests prior. But perhaps that's because of Julia 1.7 changing the random numbers, rather than a complication of your PR.

PR #437, opened 2021-11-18
Update to Gurobi 9.5

odow:
The script to update the hashes is
```Julia
run(`wget https://packages.gurobi.com/9.5/gurobi9.5.0_linux64.tar.gz`)
using Tar, Inflate, SHA
filename = "gurobi9.5.0_linux64.tar.gz"
println("sha256: ", bytes2hex(open(sha256, filename)))
println("git-tree-sha1: ", Tar.tree_hash(IOBuffer(inflate_gzip(filename))))
```

PR #456, opened 2022-02-17
WIP: Add a silent option to environment constructor

simonbowly:
Possible fix for the stray output you are seeing with 9.5 as noted in #437

1. Printing the license owners's email address: this can only be silenced by creating an empty environment, setting outputflag to zero, then starting the environment.
2. Stray printing of NonConvex parameters. This would also be silenced by this strategy, however setting OutputFlag first should have also achieved this. Is there something different done by full_bridge_optimizer that might be interfering with the Silent setting? It seems to be only these tests which print parameters.

### DIFF
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -97,10 +97,14 @@ mutable struct Env
     finalize_called::Bool
     attached_models::Int
 
-    function Env()
+    function Env(; output_flag::Int = 1)
         a = Ref{Ptr{Cvoid}}()
-        ret = GRBloadenv(a, C_NULL)
+        ret = GRBemptyenv(a)
         env = new(a[], false, 0)
+        _check_ret(env, ret)
+        ret = GRBsetintparam(env.ptr_env, GRB_INT_PAR_OUTPUTFLAG, output_flag)
+        _check_ret(env, ret)
+        ret = GRBstartenv(env.ptr_env) 
         finalizer(env) do e
             e.finalize_called = true
             if e.attached_models == 0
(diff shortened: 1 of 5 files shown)


src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -97,10 +97,15 @@ mutable struct Env
     finalize_called::Bool
     attached_models::Int
 
-    function Env()
+    function Env(::MOI.Silent, flag::Bool)
```
odow, line 100: Why not just `Env(; output_flag::Int = 1)`?

odow (commented):
CI failed because you pushed to your `simonbowly` fork. If you want to run the CI, you can push to the `jump-dev` remote:

```
git add remote jump-dev https://github.com/jump-dev/Gurobi.jl.git
git push -u jump-dev silent-env
```

odow:
Ah, I may have found the issue: 
https://github.com/jump-dev/Gurobi.jl/blob/1648d6e316835b492020bac9ee4e22bf02e5102b/src/MOI_wrapper/MOI_wrapper.jl#L382-L388

Looks like we set silent _after_ we change other parameters 😥

simonbowly:
Ah, nice one! And that's called in every constructor?

odow:
If you create a new `Optimizer`, or call` empty!`, yes. Let's just switch the two statements.

simonbowly:
I still get a couple of outputs when running the tests with that change. Are there a few Silent=False settings hanging about?

```
Set parameter OutputFlag to value 1
Set parameter OutputFlag to value 1
Set parameter NonConvex to value 2
Set parameter DualReductions to value 0
Set parameter InfUnbdInfo to value 1
Set parameter QCPDual to value 1
```

simonbowly:
CI test run for just the parameter ordering change is here - https://github.com/simonbowly/Gurobi.jl/runs/5240766098?check_suite_focus=true

This cleans up everything except the licensing messages, second commit is needed to do that.

PR #457, opened 2022-02-18
Silence test outputs

simonbowly:
Continuing from #456


odow:
I rebased because I saw this was a few commits behind master. But it looks like the prints are coming from 
https://github.com/jump-dev/Gurobi.jl/blob/1648d6e316835b492020bac9ee4e22bf02e5102b/test/MOI/MOI_wrapper.jl#L180-L217
This creates new optimizers and doesn't set silent, so we're seeing the expected behavior.

odow (approved):

odow:
Thanks Simon.

PR #465, opened 2022-04-06
allow memory limit set for gurobi 9.5.0 and higher

maxstb:


src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -97,13 +97,17 @@ mutable struct Env
     finalize_called::Bool
     attached_models::Int
 
-    function Env(; output_flag::Int = 1)
+    function Env(; output_flag::Int = 1, memory_limit::Number = 1.0e100)
```
odow, line 100:
```suggestion
    function Env(;
        output_flag::Int = 1,
        memory_limit::Union{Nothing,Real} = nothing,
    )
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -97,13 +97,17 @@ mutable struct Env
...
         a = Ref{Ptr{Cvoid}}()
         ret = GRBemptyenv(a)
         env = new(a[], false, 0)
         _check_ret(env, ret)
         ret = GRBsetintparam(env.ptr_env, GRB_INT_PAR_OUTPUTFLAG, output_flag)
         _check_ret(env, ret)
+        if _GUROBI_VERSION >= v"9.5.0"
```
odow, line 107:
```suggestion
        if _GUROBI_VERSION >= v"9.5.0" && memory_limit !== nothing
```

odow (commented):
Are there any other parameters that need to be set before the environment is started? If so, we should probably make a general fix, rather than these two explicitly parameters.

src/MOI_wrapper/MOI_wrapper.jl
odow: What about this option? Then we only set it if the user passes something, which guards against this default changing.

maxstb:
> Are there any other parameters that need to be set before the environment is started? If so, we should probably make a general fix, rather than these two explicitly parameters.

I don't know about other parameters yet.

odow (approved):
This should do for now then. Don't worry about the CI failure, you don't have access to the license.

maxstb:
Perfect, Thanks :)

PR #484, opened 2022-07-05
Add support for Absolute/Relative gap tolerances

mtanneau:
I went for the easy way: fall back to the `RawOptimizerAttribute`

___

EDIT: I have updated the PR to follow the conclusions of https://github.com/jump-dev/MathOptInterface.jl/issues/1936
At the moment, the present branch only works with `MOI#master`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,7 +9,7 @@ Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 
 [compat]
-MathOptInterface = "1.6.1"
+MathOptInterface = "1.7"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -9,7 +9,7 @@ Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 
 [compat]
-MathOptInterface = "1"
+MathOptInterface = "1.6.1"
 julia = "1.6"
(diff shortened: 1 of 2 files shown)


odow:
Hmmmmmm. I don't know if we've ever classified `RelativeGap` as a settable attribute. It's not solver-independent because every solver implements it differently.

mtanneau:
> It's not solver-independent because every solver implements it differently.

That's already the case for the getter part, and there's a warning for that in the [MOI docs](https://jump.dev/MathOptInterface.jl/stable/reference/models/#MathOptInterface.RelativeGap).

odow:
Yeah so the difference is:

 * `ModelAttribute`s are typically part of the solution information. In this case, `is_set_by_optimize(::RelativeGap) = true`:  https://github.com/jump-dev/MathOptInterface.jl/blob/bcc869f09bc3916683c6d296237ba5bfe25aa6f9/src/attributes.jl#L1968
 * `OptimizerAttribute`s are things like solver parameters.

So `RelativeGap` is really "what is the gap of the final solution," rather than the solver parameter that controls the relative gap. You could see there might be a difference, when if you `set` `RelativeGap`, what you `get` back might be smaller than that.

mtanneau:
I'll move the high-level discussion to the MOI issue. I'll update this PR based on whatever is decided there.

mtanneau:
I'll update this PR to use `MOI.AbsoluteGapTolerance` and `MOI.RelativeGapTolerance`

mtanneau:
Bump. MOI v1.6.1 has been released, which includes the definitions of `RelativeGapTolerance` and `AbsoluteGapTolerance` as `OptimizerAttribute`s.

I have updated the compat entry for MOI to `1.6.1`.

PR #488, opened 2022-11-09
Support Gurobi 10.0

metab0t:
The Clang.jl script uses an old version, so I downloaded Julia 1.5 to generate the 10.0 bindings. Maybe we should update to latest Clang.jl in the future🤔

Tests all pass locally.


odow (approved):
Looks good. I'll merge this then update the binaries.

odow:
Was this for a beta release?

metab0t:
It has not been officially announced but the download link for 10.0 is available (by inferring the URL pattern).

odow:
> It has not been officially announced but the download link for 10.0 is available

Haha. I'll wait for the official announcement to update the binaries.

PR #495, opened 2023-01-01
Change heuristic callback status to UNKNOWN instead of REJECTED

### LINKED ISSUE #494: User-provided heuristic solutions incorrectly report being rejected
Gurobi can [accept user provided heuristic solutions](https://www.gurobi.com/documentation/10.0/refman/py_model_cbsetsolution.html) during "MIP", "MIPNODE" and "MIPSOL". When in "MIPNODE", the solutions are used immediately, otherwise they are [stored to be used later](https://www.gurobi.com/documentation/10.0/refman/py_model_cbusesolution.html).

When setting solutions during "MIPSOL", Gurobi.jl incorrectly returns `HEURISTIC_SOLUTION_REJECTED` despite being used later. It could be less confusing to return `HEURISTIC_SOLUTION_UNKNOWN` instead.

Example code:

```
using JuMP
using Gurobi

m = direct_model(Gurobi.Optimizer())

@variable m x >= 0
@variable m xx
@variable m y >= 0
@variable m yy
@constraint m xx == x * x
@constraint m yy == y * y
@constraint m x + y >= 1

@objective m Min xx * x - x * x + 2 * x * y - y * y + yy * y

cb_calls = Cint[]
function my_callback_function(cb_data, cb_where::Cint)
        push!(cb_calls, cb_where)

        if cb_where != GRB_CB_MIPSOL
                return
        end

	vars = [x, y]
	vals = [0.997, 0.003]

        acc = MOI.submit(m, MOI.HeuristicSolution(cb_data), vars, vals)

	# Should be always accepted, but it never is unless queried at MIPNODE.
        if acc == MOI.HEURISTIC_SOLUTION_ACCEPTED
                throw("yay")
        end
end

MOI.set(m, MOI.RawOptimizerAttribute("NonConvex"), 2)
MOI.set(m, Gurobi.CallbackFunction(), my_callback_function)
optimize!(m)
```
Solutions are clearly being accepted, Gurobi.jl just can not make the determination yet at query time.
```
Gurobi Optimizer version 10.0.0 build v10.0.0rc2 (linux64)

CPU model: Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz, instruction set [SSE2|AVX|AVX2]
Thread count: 2 physical cores, 4 logical processors, using up to 4 threads

Optimize a model with 1 rows, 4 columns and 2 nonzeros
Model fingerprint: 0x8c47dbf8
Model has 5 quadratic objective terms
Model has 2 quadratic constraints
Coefficient statistics:
  Matrix range     [1e+00, 1e+00]
  QMatrix range    [1e+00, 1e+00]
  QLMatrix range   [1e+00, 1e+00]
  Objective range  [0e+00, 0e+00]
  QObjective range [2e+00, 4e+00]
  RHS range        [1e+00, 1e+00]

Continuous model is non-convex -- solving as a MIP

Presolve time: 0.00s
Presolved: 14 rows, 9 columns, 25 nonzeros
Presolved model has 5 bilinear constraint(s)
Variable types: 9 continuous, 0 integer (0 binary)

Root relaxation: unbounded, 2 iterations, 0.00 seconds (0.00 work units)

    Nodes    |    Current Node    |     Objective Bounds      |     Work
 Expl Unexpl |  Obj  Depth IntInf | Incumbent    BestBd   Gap | It/Node Time

     0     0  postponed    0               -          -      -     -    0s
     0     0  postponed    0               -          -      -     -    0s
H    0     0                       0.2500000          -      -     -    0s
H    0     2                       0.0029910          -      -     -    0s
H    0     2                       0.0029910          -      -     -    0s
     0     2  postponed    0         0.00299          -      -     -    0s
H   40     5                       0.0000000  -38.14697      -   0.8    0s
*   51     5              24      -0.0000000  -15.25879      -   1.0    0s

Explored 61 nodes (74 simplex iterations) in 0.01 seconds (0.00 work units)
Thread count was 4 (of 4 available processors)

Solution count 3: -1.96408e-08 0.00299097 0.25 

Optimal solution found (tolerance 1.00e-04)
Best objective 2.337765143690e-07, best bound -1.964078566452e-08, gap -

User-callback calls 306, time in user-callback 0.03 sec
```
Versions:
Gurobi version 10
Gurobi.jl v0.11.4

odow:
Closes #494

@votroto is this what you were after?

### DIFF
--- a/test/MOI/MOI_callbacks.jl
+++ b/test/MOI/MOI_callbacks.jl
@@ -320,7 +320,7 @@ function test_heuristic_callback()
                 MOI.HeuristicSolution(cb_data),
                 x,
                 ceil.(x_vals),
-            ) == MOI.HEURISTIC_SOLUTION_UNKNOWN
+            ) == MOI.HEURISTIC_SOLUTION_REJECTED
                 solution_rejected = true
             end
         end,
@@ -496,12 +496,18 @@ function test_CallbackFunction_callback_HeuristicSolution()
     model, x, item_weights = callback_knapsack_model()
     solution_accepted = false
     solution_rejected = false
+    solution_unknown = false
     cb_calls = Int32[]
     MOI.set(
         model,
         Gurobi.CallbackFunction(),
         (cb_data, cb_where) -> begin
             push!(cb_calls, cb_where)
+            if cb_where == Gurobi.GRB_CB_MIP
+                attr = MOI.HeuristicSolution(cb_data)
+                status = MOI.submit(model, attr, x, fill(2.0, length(x)))
+                solution_unknown |= (status == MOI.HEURISTIC_SOLUTION_UNKNOWN)
+            end
             if cb_where != Gurobi.GRB_CB_MIPNODE
                 return
             end
@@ -539,6 +545,7 @@ function test_CallbackFunction_callback_HeuristicSolution()
     MOI.optimize!(model)
     @test solution_accepted
     @test solution_rejected
+    @test solution_unknown
     @test Gurobi.GRB_CB_MIPNODE in cb_calls
 end
 
--- a/test/MOI/MOI_callbacks.jl
+++ b/test/MOI/MOI_callbacks.jl
@@ -320,7 +320,7 @@ function test_heuristic_callback()
                 MOI.HeuristicSolution(cb_data),
                 x,
                 ceil.(x_vals),
-            ) == MOI.HEURISTIC_SOLUTION_REJECTED
+            ) == MOI.HEURISTIC_SOLUTION_UNKNOWN
                 solution_rejected = true
             end
         end,
--- a/test/MOI/MOI_callbacks.jl
+++ b/test/MOI/MOI_callbacks.jl
@@ -506,7 +506,8 @@ function test_CallbackFunction_callback_HeuristicSolution()
             if cb_where == Gurobi.GRB_CB_MIP
                 attr = MOI.HeuristicSolution(cb_data)
                 status = MOI.submit(model, attr, x, fill(2.0, length(x)))
-                solution_unknown |= (status == MOI.HEURISTIC_SOLUTION_UNKNOWN)
+                solution_unknown |=
+                    (status == MOI.HEURISTIC_SOLUTION_UNKNOWN)
             end
             if cb_where != Gurobi.GRB_CB_MIPNODE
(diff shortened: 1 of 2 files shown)


votroto:
Thanks! Well, possibly...
It's better, but as i understand it when the "where" parameter is "MIPNODE", interpreting an infinite value as rejection is valid. However, since version 9.5 gurobi now allows users to provide heuristic solutions also during "MIP" and "MIPSOL". Users no longer have to hold onto solutions computed from a feasible solution during MIPSOL (or MIP) until the next MIPNODE comes around, gurobi now does it internally. This is why i belive it is reporting infinity (solution not found) in those cases too -- not because it was rejected, but it was not even tried yet. 

I have a question on their support forum asking for clarification and will report with the result.

odow:
Updated, with a test.

test/MOI/MOI_callbacks.jl
```diff
@@ -496,12 +496,18 @@ function test_CallbackFunction_callback_HeuristicSolution()
...
     cb_calls = Int32[]
     MOI.set(
         model,
         Gurobi.CallbackFunction(),
         (cb_data, cb_where) -> begin
             push!(cb_calls, cb_where)
+            if cb_where == Gurobi.GRB_CB_MIP
+                attr = MOI.HeuristicSolution(cb_data)
+                status = MOI.submit(model, attr, x, fill(2.0, length(x)))
+                solution_unknown |= (status == MOI.HEURISTIC_SOLUTION_UNKNOWN)
```
odow, line 509:
```suggestion
                solution_unknown |=
                    (status == MOI.HEURISTIC_SOLUTION_UNKNOWN)
```

votroto:
Nice, thank you! You were very quick.
Yes, Gurobi staff [confirmed this](https://support.gurobi.com/hc/en-us/community/posts/11822617126929/comments/11852962775569). Here is the report as promissed:

> Your interpretation is right. In the MIP and MIPSOL callback, the processing of the provided solution is delayed until the next possible point in the solution process, while in the MIPNODE callback the solution can be processed immediately. Because the solution is not checked at this point in the MIPSOL and MIP callback, the obj value is infinity. But still, the solution should be used later provided that Gurobi does not find a better solution in the meanwhile.

odow:
Thanks for following up!

PR #497, opened 2023-01-10
Add support for vector-valued objective functions

odow:
x-ref: https://github.com/jump-dev/MathOptInterface.jl/pull/2070
x-ref: https://github.com/jump-dev/JuMP.jl/issues/2099
x-ref: https://github.com/jump-dev/JuMP.jl/pull/3176

### DIFF
--- a/test/MOI/MOI_multiobjective.jl
+++ b/test/MOI/MOI_multiobjective.jl
@@ -55,7 +55,7 @@ c4: y >= 0.25
     @test MOI.get(model, Gurobi.MultiObjectivePriority(2)) == 0
 
     MOI.optimize!(model)
-    @test MOI.get(model, MOI.ObjectiveValue()) ≈ 1.5
+    @test MOI.get(model, MOI.ObjectiveValue()) ≈ [1.5, 2.0]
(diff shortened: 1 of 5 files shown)


test/MOI/MOI_multiobjective.jl
```diff
@@ -55,7 +55,7 @@ c4: y >= 0.25
     @test MOI.get(model, Gurobi.MultiObjectivePriority(2)) == 0
 
     MOI.optimize!(model)
-    @test MOI.get(model, MOI.ObjectiveValue()) ≈ 1.5
+    @test MOI.get(model, MOI.ObjectiveValue()) ≈ [1.5, 2.0]
```
odow, line 58: **This is a breaking change** (albeit, a minor one for someone using the multi-objective interface to Gurobi)

odow: @blegat: this is a good enough time as any to release a Gurobi v1.0.0?

blegat: I don't think we plan any more breaking changes so I don't see any objection

odow: Okay. I'll merge and then make a few other changes before tagging Gurobi v1.0

PR #505, opened 2023-02-14
Prep for v1.0.0

odow:
For most users this should be non-breaking. 

The only breakage is https://github.com/jump-dev/Gurobi.jl/pull/497, in which querying `ObjectiveValue` of a multi-objective optimization problem now returns a vector instead of the objective value of the first scalar objective.

This also has a minimal impact on the Pkg ecosystem because there is only one package with a dependency: https://juliahub.com/ui/Packages/Gurobi/do9v6/0.11.5 (cc @JoshKImperial for https://github.com/infrasenselabs/RoundAndSwap.jl)


odow:
I'll leave this open for a few days before merging for any objections.

PR #523, opened 2023-10-19
First constraint index to 1 instead of 2

blegat:
Closes https://github.com/jump-dev/Gurobi.jl/issues/522


odow (approved):
I added a small test

PR #528, opened 2023-10-30
Update headers for Gurobi v11 release

simonbowly:
- Add auto-generated files based on Gurobi version 11 beta
- Adjust aliases and paths


scripts/Project.toml
```diff
@@ -2,4 +2,4 @@
 Clang = "40e3b903-d033-50b4-a0cc-940c62c95e31"
 
 [compat]
-Clang = "0.12"
\ No newline at end of file
+Clang = "0.12"
```
odow, line 5:
```suggestion
Clang = "0.17"
```
I didn't test, but try with this. It's what I needed for https://github.com/jump-dev/HiGHS.jl/pull/175

odow:
You'll also need to push to the `jump-dev` remote to be able to access the CI secrets.

scripts/Project.toml
simonbowly: I tried this, without success so far. It's a different API for clang.jl, and produces a different file structure.

Do we need a generated Gurobi_JLL package as well to migrate to the new version?

odow: Oh, we'd need to update the script as well: https://github.com/jump-dev/HiGHS.jl/tree/master/gen. No need for a JLL. And no real need to change. It's a low priority and you seem to have rebuild the wrapper okay.

odow:
Once 11 is released we'll also need to update https://github.com/jump-dev/Gurobi.jl/blob/master/Artifacts.toml

scripts/Project.toml
simonbowly: Yep, I'll open another issue for that and post my attempt at the script changes.

simonbowly:
> You'll also need to push to the `jump-dev` remote to be able to access the CI secrets.

Aha, thanks, closing this in favour of #531 then

PR #531, opened 2023-10-30
Support for Gurobi v11

simonbowly:


odow:
Can now be rebased onto `master`

simonbowly:
I've re-generated the v11 headers using the new clang version. Tests are passing locally (though we can't test in CI until the v11 packages are available). Up to you if you want to merge it now or wait for the final release.

odow (approved):
I trust you. We may as well make a release with it, and then it'll just work when people update. I know who to blame if there is an issue!

PR #532, opened 2023-10-30
Update scripts for Clang v0.17

### LINKED ISSUE #530: Update to new Clang.jl version for generating future C API wrappers
We'd need to update the script as well: https://github.com/jump-dev/HiGHS.jl/tree/master/gen. No need for a JLL. And no real need to change. It's a low priority and you seem to have rebuild the wrapper okay.

_Originally posted by @odow in https://github.com/jump-dev/Gurobi.jl/pull/528#discussion_r1376713222_

odow:
Closes #530 

@simonbowly you could try building with this. It has a few tweaks, like using `Cchar` instead of `UInt8`


odow:
Seems reasonable?

simonbowly:
Yeah, a few adjustments got it building for v10. The changes look minor, but there are some failing MOI tests on my mac that I haven't dug into yet.

simonbowly:
It looks like some constraints silently fail to be added? Perhaps the CChar change could cause this, this update changes `const GRB_LESS_EQUAL = '<'` and friends to `const GRB_LESS_EQUAL = Cchar('<')`.

src/gen100/libgrb_api.jl
```diff
@@ -53,7 +64,7 @@ function GRBgetcharattrelement(model, attrname, element, valueP)
 end
 
 function GRBsetcharattrelement(model, attrname, element, newvalue)
-    ccall((:GRBsetcharattrelement, libgurobi), Cint, (Ptr{GRBmodel}, Ptr{Cchar}, Cint, UInt8), model, attrname, element, newvalue)
+    ccall((:GRBsetcharattrelement, libgurobi), Cint, (Ptr{GRBmodel}, Ptr{Cchar}, Cint, Cchar), model, attrname, element, newvalue)
```
odow, line 67: I think the test failure is because `Cchar` is not a `UInt8` on Mac, but an `Int8`.
```julia
julia> Cchar
Int8
```
Let me take a look.

simonbowly: GH actions is showing the same failures though (on linux)?

simonbowly:
Manually adjusting the character constants (6bdd625) resolved this for me. I don't see an option to fix this in Clang.jl. I guess worst case these symbols can go in the preamble?

odow:
Hmm. It'd be nice to know why. You could just do a find-and-replace on the file once it has been built?

odow:
Otherwise just a note in the `gen.jl` script is okay. It's a very small job that happens infrequently.

simonbowly:
I might just add them to the ignore list and preamble. This is a set of symbols that isn't likely to change.

As for why it fails, you may be on the right track with the Char<->[U]Int8 mapping. Evidently this differs by processor type, not by OS? If you happen to have an ARM chip handy that might shed some light on it.

It's a bit worrying that that MOI test seems to only find an issue when it checks the number of constraints in a model though? Makes it seem like an error code check is not being done?

odow:
I didn't find why it failed yet, but it's probably something to do with lines like this:
https://github.com/jump-dev/Gurobi.jl/blob/db2d8e6f88599c5b78512fb681f53ea86cb5a22b/src/MOI_wrapper/MOI_wrapper.jl#L1363

The C API calls will convert properly, but 
```julia
julia> 'B' == Cchar('B')
false
```

PR #536, opened 2023-12-01
Add Gurobi v11 artifact for CI testing

simonbowly:


odow:
I hadn't thought about this before. But this is breaking if someone was using the automatic download of Gurobi on linux, and they update their package version without updating their license. This is probably okay for now, because we shouldn't have too many people affected and it's a simple change.

But longer term, I wonder about syncing the major version of Gurobi.jl with Gurobi, so this would become v11.0.0. The downside is then we'd only ever support the latest major version, unless we back ported fixes to Gurobi.jl v10.X.Y.

Artifacts.toml
```diff
@@ -4,5 +4,5 @@ os = "linux"
 lazy = true
 
     [[gurobilinux64.download]]
-    url = "https://packages.gurobi.com/10.0/gurobi10.0.0_linux64.tar.gz"
+    url = "https://packages.gurobi.com/11.0/gurobi11.0.0_linux64.tar.gz"
     sha256 = "91a9ce1464f5f948809fcdfbdeb55f77698ed8a6d6cfa6985295424b6ece2bd4"
```
odow, line 8: Tests still still ran `Gurobi Optimizer version 10.0.0 build v10.0.0rc2 (linux64)` because the content hash didn't change. 

We need to update the SHA etc. See https://github.com/jump-dev/Gurobi.jl/pull/490

odow:
```Julia
run(`wget https://packages.gurobi.com/11.0/gurobi11.0.0_linux64.tar.gz`)
using Tar, Inflate, SHA
filename = "gurobi11.0.0_linux64.tar.gz"

julia> println("sha256: ", bytes2hex(open(sha256, filename)))
sha256: 6a1ec7499b230aea0542bc893bf0642ae8ce983dd5ef0c37cb3a253d827ce634

julia> println("git-tree-sha1: ", Tar.tree_hash(IOBuffer(inflate_gzip(filename))))
git-tree-sha1: 79f6322f611b73845e645125d5a54f17b16f6428
```

odow, line 8:
```suggestion
    sha256 = "6a1ec7499b230aea0542bc893bf0642ae8ce983dd5ef0c37cb3a253d827ce634"
```

simonbowly: Thanks, that was a fairly obvious oversight on my part ...

simonbowly:
I also missed the fallback path for the artifact in the library loader. That's updated now, so I think things are working as expected this time.

odow:
Confirm that I see `Gurobi Optimizer version 11.0.0 build v11.0.0rc2 (linux64 - "Ubuntu 22.04.3 LTS")`

PR #543, opened 2024-02-15
WIP: force update after deleting constraint

odow:
Part of debugging https://support.gurobi.com/hc/en-us/requests/66420


odow:
Closing because this didn't work

PR #544, opened 2024-02-15
Check all return codes

odow:


odow:
This doesn't fix my segfault, but it is an improvement.

PR #548, opened 2024-03-05
Check feasibility of solutions in PrimalStatus and DualStatus

### LINKED ISSUE #546: Strange reporting with respect to feasibility
I was solving a problem using JuMP with the Gurobi backend and noticed that the reporting of the termination status is odd. As an example, I get the following output after starting the solution process:

```
Gurobi Optimizer version 11.0.0 build v11.0.0rc2 (linux64 - "Ubuntu 22.04.3 LTS")                                                                                            
                  Objective                Residual                                                                                                                          
Iter       Primal          Dual         Primal    Dual     Compl     Time                                                                                                    
   0   1.34827307e+08  0.00000000e+00  1.50e-06 1.03e+02  8.35e+03     7s   
...
  51   1.54690402e+06  1.53764581e+06  3.73e+08 4.93e-04  1.12e-01    22s                                                                                                    

Barrier performed 51 iterations in 21.86 seconds (57.71 work units)
Sub-optimal termination - objective 1.54690402e+06

```

As you can see, the barrier method produces a solution that is completely unusable with a primal residuum of `1e8`, so the solution is *not* feasible. However, `@info solution_summary(model)` produces the output

```
┌ Info: * Solver : Gurobi                                                             
│                                              
│ * Status                                  
│   Result count       : 1                                                            
│   Termination status : LOCALLY_SOLVED                             
│   Message from the solver:                
│   "Unable to satisfy optimality tolerances; a sub-optimal solution is available."
│                                           
│ * Candidate solution (result jump-dev/JuMP.jl#1)          
│   Primal status      : FEASIBLE_POINT     
│   Dual status        : NO_SOLUTION        
│   Objective value    : 1.54690e+06        
│                                           
│ * Work counters                                                   
│   Solve time (sec)   : 2.18862e+01        
│   Barrier iterations : 51                 
└   Node count         : 0          
```

The primal status message indicates that the solution is feasible, which it is not. Sub-optimal solutions may be primally / dually feasible depending on the specific method (for instance with the primal / dual simplex), but the central path followed by the interior point method generally only yields feasible points in the limit (at least as far as I know).

I don't know how the reporting works of feasibility works inside JuMP and which solvers are affected, but it is a bit problematic if a user expects a feasible solution when none exists.

odow:
Closes #546 

This is one option, but I don't really like it, because it relies on magic constants. Is there a way for Gurobi to tell us whether the problem is feasible according to any parameters the user has set?

### DIFF
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -2834,6 +2834,10 @@ function MOI.get(model::Optimizer, attr::MOI.PrimalStatus)
     if !(1 <= attr.result_index <= valueP[])
         return MOI.NO_SOLUTION
     end
+    if term in (MOI.OPTIMAL, MOI.SOLUTION_LIMIT)
+        return MOI.FEASIBLE_POINT
+    end
+    # Feasibility of solution is unknown. Check for violations.
     doubleP = Ref{Cdouble}()
     ret = GRBgetdblattr(model, "MaxVio", doubleP)
     _check_ret(model, ret)
@@ -2890,6 +2894,10 @@ function MOI.get(model::Optimizer, attr::MOI.DualStatus)
     if ret != 0  # Something went wrong
         return MOI.NO_SOLUTION
     end
+    if term in (MOI.OPTIMAL, MOI.SOLUTION_LIMIT)
+        return MOI.FEASIBLE_POINT
+    end
+    # Feasibility of solution is unknown. Check for violations.
     ret = GRBgetdblattr(model, "MaxVio", doubleP)
     _check_ret(model, ret)
     if doubleP[] < 1e-8
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -2831,8 +2831,19 @@ function MOI.get(model::Optimizer, attr::MOI.PrimalStatus)
     valueP = Ref{Cint}()
     ret = GRBgetintattr(model, "SolCount", valueP)
     _check_ret(model, ret)
-    return 1 <= attr.result_index <= valueP[] ? MOI.FEASIBLE_POINT :
-           MOI.NO_SOLUTION
+    if !(1 <= attr.result_index <= valueP[])
+        return MOI.NO_SOLUTION
+    end
+    doubleP = Ref{Cdouble}()
+    ret = GRBgetdblattr(model, "MaxVio", doubleP)
+    _check_ret(model, ret)
+    if doubleP[] < 1e-8
+        return MOI.FEASIBLE_POINT
+    elseif doubleP[] < 1e-6
+        return MOI.NEARLY_FEASIBLE_POINT
+    else
+        return MOI.INFEASIBLE_POINT
+    end
 end
 
 function _has_dual_ray(model::Optimizer)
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -2831,8 +2831,20 @@ function MOI.get(model::Optimizer, attr::MOI.PrimalStatus)
...
+        return MOI.FEASIBLE_POINT
+    end
+    # Feasibility of solution is unknown. Check for violations.
+    doubleP = Ref{Cdouble}()
+    ret = GRBgetdblattr(model, "MaxVio", doubleP)
+    _check_ret(model, ret)
+    if doubleP[] < 1e-8
+        return MOI.FEASIBLE_POINT
+    end
+    return MOI.UNKNOWN_RESULT_STATUS
```
blegat, line 2847: This should be `INFEASIBLE_POINT`

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -2831,8 +2831,20 @@ function MOI.get(model::Optimizer, attr::MOI.PrimalStatus)
...
+        return MOI.NO_SOLUTION
+    end
+    if term in (MOI.OPTIMAL, MOI.SOLUTION_LIMIT)
+        return MOI.FEASIBLE_POINT
+    end
+    # Feasibility of solution is unknown. Check for violations.
+    doubleP = Ref{Cdouble}()
+    ret = GRBgetdblattr(model, "MaxVio", doubleP)
+    _check_ret(model, ret)
+    if doubleP[] < 1e-8
```
blegat, line 2844: If we make this `1e-8` a parameter that the user can modify then I think it's fine

blegat:
The other option is to just return `UNKNOWN_RESULT_STATUS`

odow:
Always returning `UNKNOWN_RESULT_STATUS` will break returning a feasible point on a time limit.

src/MOI_wrapper/MOI_wrapper.jl
odow: I don't really like this, because it won't respect any parameters that are set by the user.

odow:
Support confirm there is nothing in the API for this, so I guess we're stuck with a magic constant.

simonbowly:
> Support confirm there is nothing in the API for this, so I guess we're stuck with a magic constant.

Why compare with a hard-coded constant instead of the tolerance parameters set by the user?

odow:
>  instead of the tolerance parameters set by the user?

I mean, we could. But then we'd need to loop through all tolerances and all violations and account for any nasty differences between algorithms etc (without being able to know what they are). It'd turn into the function I want in the C API that you're running up the flagpole...

simonbowly:
Sure, this is more to point out that the default tolerances are 10^-6, so you can expect a lot of false conclusions with a hard coded 10^-8. But now I see @blegat's suggestion above for a jump-side parameter.

odow:
Yeah we wouldn't actually hard-code. And `1e-8` is just a placeholder.

odow:
Okay, how's this. I assumed that is sufficient to check only `ConstrVio`, `BoundVio`, and `IntVio`.

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -2869,18 +2920,22 @@ function MOI.get(model::Optimizer, attr::MOI.DualStatus)
...
         return MOI.NO_SOLUTION
     end
     # But unfortunately, even if SolCount is 1, sometimes a dual solution is not
     # available. The only way to check this is to query directly.
-    valueP = Ref{Cdouble}()
-    ret = GRBgetdblattrelement(model, "RC", 0, valueP)
-    if ret == 0
-        return MOI.FEASIBLE_POINT
-    else  # Something went wrong
+    doubleP = Ref{Cdouble}()
```
odow, line 2928:
```suggestion
    valueP = Ref{Cdouble}()
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -2869,18 +2920,22 @@ function MOI.get(model::Optimizer, attr::MOI.DualStatus)
...
+    ret = GRBgetdblattrelement(model, "RC", 0, doubleP)
```
odow, line 2929:
```suggestion
    ret = GRBgetdblattrelement(model, "RC", 0, valueP)
```

PR #550, opened 2024-03-12
WIP. Refactor GRBupdatemodel calls

torressa:
WIP.
The idea is to only call `GRBupdatemodel` before `GRBget*`.
Some tests are failing still: 
- [ ] `test_Buffered_deletion_test`, `test_modify_after_delete*` (indexing should also be updated with delete? This is not expected from our other APIs)

Fixes https://github.com/jump-dev/Gurobi.jl/issues/516

### DIFF
--- a/src/MOI_wrapper/MOI_multi_objective.jl
+++ b/src/MOI_wrapper/MOI_multi_objective.jl
@@ -36,7 +36,6 @@ function MOI.set(
         obj[column] += term.coefficient
     end
     indices, coefficients = _indices_and_coefficients(model, f)
-    _update_if_necessary(model)
     ret = GRBsetobjectiven(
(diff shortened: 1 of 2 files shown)


odow:
CI will fail because you don't have access to the CI secrets. I'll send you an invite, and then you can push to the `jump-dev` remote and get access to CI.

odow:
I remember when writing parts of this that I got _very_ confused as to when a call to GRBupdate was necessary. So I defaulted to requiring an update anytime the model was modified, and ensuring update had been called before querying any data about the model.

torressa:
Thanks Oscar! I've managed to get the CI to run. I'll continue in the other PR. Sorry for the mess.

Issue #516, opened 2023-08-16
Refactor constraint deletions and GRBupdatemodel


odow:
The MOI wrapper eagerly updates the row  indices of constraints in a model upon deletion: 

https://github.com/jump-dev/Gurobi.jl/blob/38464c27b96e9818257b22c61d2458890957d0fe/src/MOI_wrapper/MOI_wrapper.jl#L2115-L2129

We should instead do what we do for variables and lazily delay the update until the next call to `GRBupdatemodel`:

https://github.com/jump-dev/Gurobi.jl/blob/38464c27b96e9818257b22c61d2458890957d0fe/src/MOI_wrapper/MOI_wrapper.jl#L981-L986

https://github.com/jump-dev/Gurobi.jl/blob/38464c27b96e9818257b22c61d2458890957d0fe/src/MOI_wrapper/MOI_wrapper.jl#L520-L534

This also explains why we didn't trigger #515 for variable deletion: https://github.com/jump-dev/Gurobi.jl/pull/515#discussion_r1294262326

odow:
Potentially related to https://support.gurobi.com/hc/en-us/requests/66420

I've been trying to track down a segfault for dkee331. I don't think I can share the model because it contains some industry data.

```
julia> include("SDDP-models.jl")


*** Start recording in file recording012.grbr

Change of opaque parameter Username not recorded
Change of opaque parameter GURO_PAR_LICFILE not recorded
-------------------------------------------------------------------
         SDDP.jl (c) Oscar Dowson and contributors, 2017-23
-------------------------------------------------------------------
problem
  nodes           : 12
  state variables : 15
  scenarios       : 7.43008e+11
  existing cuts   : false
options
  solver          : serial mode
  risk measure    : SDDP.Expectation()
  sampling scheme : SDDP.InSampleMonteCarlo
subproblem structure
  VariableRef                             : [42, 47]
  AffExpr in MOI.EqualTo{Float64}         : [15, 15]
  AffExpr in MOI.LessThan{Float64}        : [11, 11]
  VariableRef in MOI.GreaterThan{Float64} : [16, 17]
  VariableRef in MOI.LessThan{Float64}    : [1, 1]
numerical stability report
  matrix range     [2e-03, 2e+01]
  objective range  [0e+00, 0e+00]
  bounds range     [1e+07, 1e+07]
  rhs range        [7e-02, 3e+03]
-------------------------------------------------------------------
[...]

PR #552, opened 2024-03-13
Refactor GRBupdatemodel calls

torressa:
Motivated by the large performance difference in [Discourse: Problem with JuMP + Gurobi.jl](https://discourse.julialang.org/t/problem-with-jump-gurobi-jl/111505). 

The idea is to only call `GRBupdatemodel` before `GRBget*` only. I've tried to adhere to calling `_update_if_required` at the start of the function or at the end. Additionally, adding the `force` argument to this function avoids having to use `_require_update`.

Some tests are failing still:

- [x] `test_Buffered_deletion_test`, `test_modify_after_delete*` (indexing should also be updated with delete? This is not expected from our other APIs)
- [x] `test_linear_SOS1_integration`
- [x] `test_set_basis`. When setting + getting `[V|C]Basis` we need to update beforehand. Doing this on the `MOI.set/get` leads to a lot of warning messages and (potentially) performance issues.

Work for the update part of https://github.com/jump-dev/Gurobi.jl/issues/516

### DIFF
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -1038,6 +1038,7 @@ function MOI.delete(model::Optimizer, indices::Vector{<:MOI.VariableIndex})
     # We throw away name_to_constraint_index so we will rebuild VariableIndex
     # constraint names without v.
     model.name_to_constraint_index = nothing
+    # _require_update(model, model_change = true)
     _update_if_necessary(model, force = true)
     return
 end
@@ -1055,7 +1056,7 @@ function MOI.delete(model::Optimizer, v::MOI.VariableIndex)
     # We throw away name_to_constraint_index so we will rebuild VariableIndex
     # constraint names without v.
     model.name_to_constraint_index = nothing
-    _update_if_necessary(model, force = true)
+    _require_update(model, model_change = true)
     return
 end
 
@@ -1565,7 +1566,7 @@ function MOI.delete(
     end
     info.upper_bound_if_bounded = NaN
     model.name_to_constraint_index = nothing
-    _update_if_necessary(model)
+    _require_update(model, attribute_change = true)
     return
 end
 
@@ -1680,7 +1681,7 @@ function MOI.delete(
     end
     info.lower_bound_if_bounded = NaN
     model.name_to_constraint_index = nothing
-    _update_if_necessary(model)
+    _require_update(model, attribute_change = true)
     return
 end
 
@@ -1696,7 +1697,7 @@ function MOI.delete(
     info.bound = _NONE
     info.upper_bound_if_bounded = info.lower_bound_if_bounded = NaN
     model.name_to_constraint_index = nothing
-    _update_if_necessary(model)
+    _require_update(model, attribute_change = true)
     return
 end
 
@@ -1712,7 +1713,7 @@ function MOI.delete(
     info.bound = _NONE
     info.upper_bound_if_bounded = info.lower_bound_if_bounded = NaN
     model.name_to_constraint_index = nothing
-    _update_if_necessary(model)
+    _require_update(model, attribute_change = true)
     return
 end
 
@@ -1860,7 +1861,7 @@ function MOI.delete(
     _check_ret(model, ret)
     info.type = GRB_CONTINUOUS
     model.name_to_constraint_index = nothing
-    _update_if_necessary(model)
+    _require_update(model, attribute_change = true)
     return
 end
 
@@ -1898,7 +1899,7 @@ function MOI.delete(
     _check_ret(model, ret)
     info.type = GRB_CONTINUOUS
     model.name_to_constraint_index = nothing
-    _update_if_necessary(model)
+    _require_update(model, attribute_change = true)
     return
 end
 
@@ -1946,7 +1947,7 @@ function MOI.delete(
     _check_ret(model, ret)
     info.type = GRB_CONTINUOUS
     model.name_to_constraint_index = nothing
-    _update_if_necessary(model)
+    _require_update(model, attribute_change = true)
     return
 end
 
@@ -2000,7 +2001,7 @@ function MOI.delete(
     _check_ret(model, ret)
     info.type = GRB_CONTINUOUS
     model.name_to_constraint_index = nothing
-    _update_if_necessary(model)
+    _require_update(model, attribute_change = true)
     return
 end
 
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -3715,6 +3715,7 @@ function MOI.set(
     c::MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64},S},
     bs::MOI.BasisStatusCode,
 ) where {S<:_SCALAR_SETS}
+    _update_if_necessary(model)
     row = _info(model, c).row
     valueP::Cint = (
         if bs == MOI.BASIC
@@ -3737,6 +3738,7 @@ function MOI.set(
     x::MOI.VariableIndex,
     bs::MOI.BasisStatusCode,
 )
+    _update_if_necessary(model)
     valueP = if bs == MOI.BASIC
         Cint(0)
     elseif bs == MOI.NONBASIC_AT_LOWER
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -873,6 +873,7 @@ function column(
 )
     return _info(model, x).column
 end
+
 """
     c_column(
         model::Optimizer,
@@ -1000,7 +1001,6 @@ function MOI.is_valid(model::Optimizer, v::MOI.VariableIndex)
 end
 
 function MOI.delete(model::Optimizer, indices::Vector{<:MOI.VariableIndex})
-    #_update_if_necessary(model)
     info = [_info(model, var_idx) for var_idx in indices]
     soc_idx = findfirst(e -> e.num_soc_constraints > 0, info)
     soc_idx !== nothing && throw(MOI.DeleteNotAllowed(indices[soc_idx]))
@@ -1020,7 +1020,6 @@ function MOI.delete(model::Optimizer, indices::Vector{<:MOI.VariableIndex})
 end
 
 function MOI.delete(model::Optimizer, v::MOI.VariableIndex)
-    #_update_if_necessary(model)
     info = _info(model, v)
     if info.num_soc_constraints > 0
         throw(MOI.DeleteNotAllowed(v))
@@ -1092,7 +1091,6 @@ end
 function _zero_objective(model::Optimizer)
     num_vars = length(model.variable_info)
     obj = zeros(Float64, num_vars)
-    _update_if_necessary(model)
     ret = GRBdelq(model)
     _check_ret(model, ret)
     ret = GRBsetdblattrarray(model, "Obj", 0, num_vars, obj)
@@ -1170,10 +1168,6 @@ function MOI.set(
     for term in f.terms
         obj[column(model, term.variable)] += term.coefficient
     end
-    # NOTE: variables added may be referred before a `_update_if_necessary`
-    # what is the problem we try to prevent below?
-    # This update is needed because we might have added some variables.
-    _update_if_necessary(model)
     ret = GRBsetdblattrarray(model, "Obj", 0, num_vars, obj)
     _check_ret(model, ret)
     ret = GRBsetdblattr(model, "ObjCon", f.constant)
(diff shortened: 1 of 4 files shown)


odow:
I don't have much advice here. It was hard to get things working without knowing the internal details. So perhaps I've papered over other mistakes that are now showing themselves.

torressa:
It's alright Oscar, I only have `test_set_basis` left. This a special case as we would need to update before `MOI.get` and `MOI.set`. This may lead to some overhead but also to a lot of warning messages. This is the Python API equivalent:

```
import gurobipy as gp

m = gp.Model()

x = m.addVar()
y = m.addVars(100)

c = m.addConstr(x + y.sum() <= 1)
m.setObjective(-x)

# Update before setting

m.update()
x.VBasis = 0

m.update()
c.CBasis = -1

for i in range(100):
    m.update()
    y[i].VBasis = -1


# Also update before getting
m.update()
print(x.VBasis)

m.optimize()
```

It might just be easier to have the user manually call `GRBupdatemodel(model)` in these cases. I will ask Simon.

odow:
> It might just be easier to have the user manually call GRBupdatemodel(model) in these cases

As a comment, ideally we should not have to do this. The goal of MOI is to abstract across solvers. In the default case, user's won't have access to an object that they can call `GRBupdate` on. 

Does the Python example throw the same warnings?

When is the warning actually printed?

I think very few people are actually setting the starting basis. But getting is a common operation.

odow:
We already update here:
https://github.com/jump-dev/Gurobi.jl/blob/4d204e22cf0f6004e9580583480835c451ebbf97/src/MOI_wrapper/MOI_wrapper.jl#L3691-L3698
So is the only thing missing an update at the start of `set`?
https://github.com/jump-dev/Gurobi.jl/blob/4d204e22cf0f6004e9580583480835c451ebbf97/src/MOI_wrapper/MOI_wrapper.jl#L3734-L3752

torressa:
> Does the Python example throw the same warnings?

Yes the warnings are from the Python example, and they will be printed every time we call update without a complete basis so once for every variable except the last one.

odow:
I'm okay if we throw the same warnings as Python. But perhaps you could fix this internally to only throw a warning if `GRBoptimize` is called without a complete basis?

odow (approved):
This doesn't change any tests and everything is still passing so LGTM.

torressa:
Simon has a better idea of how to handle this: two kinds of update flags.

- `attribute_change` in which case we don't have to call `GRBupdate` whenever setting further attributes only when `GRBget` as usual;
- `model_change` for any variable, constraint, or objective changes

Let me have another go

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1841,9 +1858,9 @@ function MOI.delete(
     ret =
         GRBsetcharattrelement(model, "VType", Cint(info.column - 1), Char('C'))
     _check_ret(model, ret)
-    _require_update(model)
     info.type = GRB_CONTINUOUS
     model.name_to_constraint_index = nothing
+    _update_if_necessary(model, force = true)
```
odow, line 1863: Are these `force` necessary if we're just changing a variable from binary to continuous?

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1664,6 +1680,7 @@ function MOI.delete(
     end
     info.lower_bound_if_bounded = NaN
     model.name_to_constraint_index = nothing
+    _update_if_necessary(model, force = true)
```
odow, line 1683:
```suggestion
    _update_if_necessary(model, attribute_change = true)
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1542,14 +1558,14 @@ function MOI.delete(
...
     _check_ret(model, ret)
-    _require_update(model)
     if info.bound == _LESS_AND_GREATER_THAN
         info.bound = _GREATER_THAN
     else
         info.bound = _NONE
     end
     info.upper_bound_if_bounded = NaN
     model.name_to_constraint_index = nothing
+    _update_if_necessary(model, force = true)
```
odow, line 1568:
```suggestion
    _update_if_necessary(model, attribute_change = true)
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1676,10 +1693,10 @@ function MOI.delete(
     _set_variable_lower_bound(model, info, -Inf)
     ret = GRBsetdblattrelement(model, "UB", Cint(info.column - 1), Inf)
     _check_ret(model, ret)
-    _require_update(model)
     info.bound = _NONE
     info.upper_bound_if_bounded = info.lower_bound_if_bounded = NaN
     model.name_to_constraint_index = nothing
+    _update_if_necessary(model, force = true)
```
odow, line 1699:
```suggestion
    _update_if_necessary(model, attribute_change = true)
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1692,10 +1709,10 @@ function MOI.delete(
...
```
odow, line 1715:
```suggestion
    _update_if_necessary(model, attribute_change = true)
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1879,9 +1896,9 @@ function MOI.delete(
     ret =
         GRBsetcharattrelement(model, "VType", Cint(info.column - 1), Char('C'))
     _check_ret(model, ret)
-    _require_update(model)
     info.type = GRB_CONTINUOUS
     model.name_to_constraint_index = nothing
+    _update_if_necessary(model, force = true)
```
odow, line 1901:
```suggestion
    _update_if_necessary(model, attribute_change = true)
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1927,9 +1944,9 @@ function MOI.delete(
     _set_variable_lower_bound(model, info, -Inf)
     ret = GRBsetdblattrelement(model, "UB", Cint(info.column - 1), Inf)
     _check_ret(model, ret)
-    _require_update(model)
     info.type = GRB_CONTINUOUS
     model.name_to_constraint_index = nothing
+    _update_if_necessary(model, force = true)
```
odow, line 1949:
```suggestion
    _update_if_necessary(model, attribute_change = true)
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1981,9 +1998,9 @@ function MOI.delete(
...
```
odow, line 2003:
```suggestion
    _update_if_necessary(model, attribute_change = true)
```

simonbowly:
@torressa it looks like you have `attribute_change=true` for some `add_constraint` calls, but `attribute_change=false` for others? Which one are you going for?

Just to spell it out to make sure we're on the same page: is the aim here to ensure that an update is done if the user is about to:

1. get() the value of some attribute for a variable/constraint that was added by the user since the last update call (because get() would otherwise throw an error)
2. get() the value of some attribute which was changed by the user since the last update call (because get() would return the old attribute value)
3. set() the value of some attribute for a variable/constraint that was added by the user since the last update call (to handle setting basis values)

?

(Side note: is there really anything to worry about with deletes? Or would JuMP error out earlier than this if the user tries to query an attribute on a deleted variable or constraint?)

torressa:
> @torressa it looks like you have attribute_change=true for some add_constraint calls, but attribute_change=false for others? Which one are you going for?

I did, yes, some `add_constraint` functions only change attributes via `GRBset*element`, so my thinking was to keep them as attribute type changes (e.g. [src/MOI_wrapper/MOI_wrapper.jl#L1897](https://github.com/jump-dev/Gurobi.jl/blob/69d2d6e70ab517751275bea0d4ec0acdb263299f/src/MOI_wrapper/MOI_wrapper.jl#L1897)).
Other `add_constraint` functions that use `GRBadd` or a combination are marked as model changes.

> Just to spell it out to make sure we're on the same page: is the aim here to ensure that an update is done if the user is about to:
> - get() the value of some attribute for a variable/constraint that was added by the user since the last update call (because get() would otherwise throw an error)
> - get() the value of some attribute which was changed by the user since the last update call (because get() would return the old attribute value)
> - set() the value of some attribute for a variable/constraint that was added by the user > since the last update call (to handle setting basis values)
> ?

Exactly, that's the aim.


> (Side note: is there really anything to worry about with deletes? Or would JuMP error out earlier than this if the user tries to query an attribute on a deleted variable or constraint?)

I'm not sure what the workflow is in practice, but in the tests, we have things like:

https://github.com/jump-dev/Gurobi.jl/blob/69d2d6e70ab517751275bea0d4ec0acdb263299f/test/MOI/MOI_wrapper.jl#L285-L303

This will fail without updating after deleting

simonbowly:
> I did, yes, some `add_constraint` functions only change attributes via `GRBset*element`, so my thinking was to keep them as attribute type changes (e.g. [src/MOI_wrapper/MOI_wrapper.jl#L1897](https://github.com/jump-dev/Gurobi.jl/blob/69d2d6e70ab517751275bea0d4ec0acdb263299f/src/MOI_wrapper/MOI_wrapper.jl#L1897)). Other `add_constraint` functions that use `GRBadd` or a combination are marked as model changes.

Got it, thanks, that makes sense now.

> set() the value of some attribute for a variable/constraint that was added by the user > since the last update call (to handle setting basis values)

Great, ok, so `_update_if_necessary(model, check_attribute_change = false)` is used only when setting basis statuses. Looks good!

> I'm not sure what the workflow is in practice, but in the tests, we have things like:
> 
> https://github.com/jump-dev/Gurobi.jl/blob/69d2d6e70ab517751275bea0d4ec0acdb263299f/test/MOI/MOI_wrapper.jl#L285-L303
> 
> This will fail without updating after deleting

At least for variables (constraints may need special handling per #516), it seems like a forced update immediately after the delete isn't needed. It should instead need `_require_update(model_change=true)` *after* a delete, so that an update is deferred until after a series of deletes. I'm not really across the details of how MOI handles index tracking on deletes though, so maybe it's better to go with the more paranoid approach for now if this fixes the performance issue you are targeting.

torressa:
> At least for variables (constraints may need special handling per https://github.com/jump-dev/Gurobi.jl/issues/516), it seems like a forced update immediately after the delete isn't needed. It should instead need _require_update(model_change=true) after a delete, so that an update is deferred until after a series of deletes. I'm not really across the details of how MOI handles index tracking on deletes though, so maybe it's better to go with the more paranoid approach for now if this fixes the performance issue you are targeting.

Thanks Simon! You are right, sorry I missed this.
One force update remains for deleting a vector of variables.

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1038,6 +1038,7 @@ function MOI.delete(model::Optimizer, indices::Vector{<:MOI.VariableIndex})
     # We throw away name_to_constraint_index so we will rebuild VariableIndex
     # constraint names without v.
     model.name_to_constraint_index = nothing
+    # _require_update(model, model_change = true)
```
torressa, line 1041: This is the only update remaining for variable deletion as is needed for the `test_Buffered_deletion_test` case (linked earlier)

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -1038,6 +1038,7 @@ function MOI.delete(model::Optimizer, indices::Vector{<:MOI.VariableIndex})
...
     _update_if_necessary(model, force = true)
```
torressa, line 1042: This is the only update remaining for variable deletion as is needed for the `test_Buffered_deletion_test` case (linked earlier)

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -503,20 +506,38 @@ end
...
-Sets the `model.needs_update` flag. Call this at the end of any mutating method.
+Sets the `model.needs_update` flag. Call this at the end of any `GRBset*` call.
 """
-function _require_update(model::Optimizer)
-    model.needs_update = true
+function _require_update(
+    model::Optimizer;
+    attribute_change::Bool = false,
+    model_change::Bool = false,
+)
```
odow, line 515:
```suggestion
)
    @assert attribute_change || model_change
```

odow: Just for sanity sake, check that we never call `_require_update(model)`

odow:
Okay. Does this solve #516? Or should we still buffer the deletions?

torressa:
I think so. I changed some deletions that don't require an immediate update (e.g. those that only change an attribute), but the others have to stay as they were to pass the tests (force an update at the end).

PR #575, opened 2024-10-02
Add support for ScalarNonlinearFunction

torressa:


src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -194,6 +196,7 @@ function MOI.add_constraint(
     model::Optimizer,
     f::MOI.ScalarNonlinearFunction,
     s::_SCALAR_SETS,
+    resvar_sense::_SCALAR_SETS = s,
```
odow, line 216: You cannot define this method.

torressa: Thanks! Yeah I'm not sure what I was thinking here 😅

odow:
```julia
begin
    model = Gurobi.Optimizer()
    config = MOI.Test.Config()
    MOI.Test.test_that_is_failing(model, config)
end
```

Although I usually find it more helpful to find the source of the test function, the copy paste it into a new file and step through to see/debug which bit is actually failing.

simonbowly:
> Although I usually find it more helpful to find the source of the test function, the copy paste it into a new file and step through to see/debug which bit is actually failing.

Got it, thanks

simonbowly:
I'm a little stuck on this one: `test_multiobjective_vector_nonlinear` is marked as failing (it's an internal error in Gurobi.jl, since this is a case we don't directly handle). But then running the test directly:

```julia
using Gurobi
import MathOptInterface as MOI

begin
    model = Gurobi.Optimizer()
    config = MOI.Test.Config()
    MOI.Test.test_multiobjective_vector_nonlinear(model, config)
end
```

exits with a RequirementUnmet:

```
Set parameter TimeLimit to value 1
Read parameters from file gurobi.env
Gurobi 12.0.0beta3 - expires 2024-11-18
ERROR: LoadError: RequirementUnmet: MOI.supports(model, MOI.ObjectiveFunction{F}())
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/MathOptInterface/HB8aJ/src/Test/Test.jl:320 [inlined]
 [2] test_multiobjective_vector_nonlinear(model::Gurobi.Optimizer, ::MathOptInterface.Test.Config{Float64})
   @ MathOptInterface.Test ~/.julia/packages/MathOptInterface/HB8aJ/src/Test/test_multiobjective.jl:237
 [3] top-level scope
   @ ~/workspaces/Gurobi.jl/tmp.jl:7
in expression starting at /Users/bowly/workspaces/Gurobi.jl/tmp.jl:4
```

as it should, since this returns `false`:

```
MOI.supports(model, MOI.ObjectiveFunction{MOI.VectorNonlinearFunction}())
```

@odow shouldn't these test be skipped since the requirements aren't met?

odow:
Yes the tests should automatically be skipped if it throws a RequirementUnmet error.

If it isn't for some reason, just add `r"^test_multiobjective_vector_nonlinear$"` to the list of excludes and move on. We can address at a later date.

You can probably also add the excludes `"vector_nonlinear"` and `"VectorNonlinearFuction"`. There are some weird issues with JuMP's bridges that I need to fix, but it hasn't been a priority because we have very few MINLP solvers.

simonbowly:
@torressa thanks for working on this! It looks in pretty good shape, I made some tweaks to fix or skip the remaining tests**. One missing thing is handling of quadratic expressions. I just added a test for that case (https://github.com/jump-dev/Gurobi.jl/pull/575/commits/086396160b69026c6fe2e7f2c230fceeac4ca595), and I should be able to implement it tomorrow (though I'm not sure why it doesn't come up in the MOI tests).

** None of the tests can actually be run here in CI until 12.0.0 is released on conda, but locally I'm now seeing just the one failure for the new test I added.

torressa:
Thanks @simonbowly for taking a look and fixing the CI 😆 

> ** None of the tests can actually be run here in CI until 12.0.0 is released on conda, but locally I'm now seeing just the one failure for the new test I added.

The same thing goes for the codecov complaints

odow:
Could we split the v12 C API patch into a new PR? That can be merged now, even without upstream released, etc.

torressa:
> Could we split the v12 C API patch into a new PR? That can be merged now, even without upstream released, etc.

Opened #582 for this. I think that's everything, all the diff will disappear once on master

odow:
I'll rebase

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    parent::Vector{Cint},
+)
+    # TODO: use type hints here instead of Any
+    stack = Vector{Tuple{Any,Cint}}([(f, Cint(-1))])
+    current_index = Cint(-1)
+
+    while length(stack) != 0
+        current_index += Cint(1)
+        s, parent_index = pop!(stack)
+
```
odow, line 150:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+                    coeff,
+                    var_index,
+                    current_index,
+                    parent_index,
+                )
+                current_index += Cint(1)
+            end
+        elseif typeof(s) == Float64
+            _add_expression_tree_constant(opcode, data, parent, s, parent_index)
+        elseif typeof(s) == Int64
```
odow, line 208:
```suggestion
        elseif s isa Int
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+                    data,
+                    parent,
+                    coeff,
+                    var_index,
+                    current_index,
+                    parent_index,
+                )
+                current_index += Cint(1)
+            end
+        elseif typeof(s) == Float64
```
odow, line 206:
```suggestion
        elseif s isa Float64
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+                parent_index,
+            )
+        elseif typeof(s) == MOI.ScalarAffineFunction{Float64}
+            if length(s.terms) > 1
+                append!(opcode, GRB_OPCODE_PLUS)
+                append!(data, -1.0)
+                append!(parent, parent_index)
+                parent_index += Cint(1)
+            end
+            if s.constant != 0.0
```
odow, line 186:
```suggestion
            if !iszero(s.constant)
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+            _add_expression_tree_variable(
+                opcode,
+                data,
+                parent,
+                1.0,
+                c_column(model, s),
+                current_index,
+                parent_index,
+            )
+        elseif typeof(s) == MOI.ScalarAffineFunction{Float64}
```
odow, line 179:
```suggestion
        elseif s isa MOI.ScalarAffineFunction{Float64}
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+                   !_check_nonlinear_constant(s)
+                append!(opcode, ret)
+                append!(data, -1.0)
+                append!(parent, parent_index)
+            end
+
+            for expr in reverse(s.args)
+                push!(stack, (expr, current_index))
+            end
+        elseif typeof(s) == MOI.VariableIndex
```
odow, line 169:
```suggestion
        elseif s isa MOI.VariableIndex
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+
+    return
+end
+
+function MOI.add_constraint(
+    model::Optimizer,
+    f::MOI.ScalarNonlinearFunction,
+    s::_SCALAR_SETS,
+)
+    opcode = Vector{Cint}()
```
odow, line 229:
```suggestion
    opcode = Cint[]
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    data = Vector{Cdouble}()
```
odow, line 230:
```suggestion
    data = Cdouble[]
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    parent = Vector{Cint}()
```
odow, line 231:
```suggestion
    parent = Cint[]
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+
```
odow, line 232:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    sense, rhs = _sense_and_rhs(s)
+
```
odow, line 234:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    _process_nonlinear(model, f, opcode, data, parent)
+
```
odow, line 236:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    # Add resultant variable
+    vi, ci = MOI.add_constrained_variable(model, s)
+    resvar_index = c_column(model, vi)
+
```
odow, line 240:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+        model,
+        C_NULL,
+        resvar_index,
+        length(opcode),
+        opcode,
+        data,
+        parent,
+    )
+    _check_ret(model, ret)
+    # GRBwrite(model, "checkjl.lp")
```
odow, line 251:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+        model.last_constraint_index,
+    )
+end
+
+function MOI.is_valid(
+    model::Optimizer,
+    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
+) where {S}
+    info = get(model.nl_constraint_info, c.value, nothing)
+    return info !== nothing && typeof(info.set) == S
```
odow, line 266:
```suggestion
    return info !== nothing && info.set isa S
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+end
+
+function MOI.delete(
+    model::Optimizer,
+    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
+) where {S}
+    info = _info(model, c)
+    _update_if_necessary(model)
+
```
odow, line 275:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    ret = GRBdelgenconstrs(model, 1, [Cint(info.row - 1)])
```
odow, line 276:
```suggestion
    ret = GRBdelgenconstrs(model, 1, Ref(Cint(info.row - 1)))
```

`[]` allocates a new vector, but really we just need to pass a pointer to something with one element.

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    _require_update(model, model_change = true)
+    return
+end
+
+function MOI.delete(
+    model::Optimizer,
+    cs::Vector{<:MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S}},
+) where {S}
+    rows_to_delete = sort!([Cint(_info(model, c).row - 1) for c in cs])
+    println(rows_to_delete)
```
odow, line 296:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+
+    # Delete resultant variables
+    resvars = [_info(model, c).resvar for c in cs]
+    MOI.delete(model, resvars)
+
+    cs_values = sort!(getfield.(cs, :value))
+    filter!(model.nl_constraint_info) do pair
+        return isempty(searchsorted(cs_values, pair.first))
+    end
+
```
odow, line 315:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    for (_, info) in model.nl_constraint_info
+        info.row -= searchsortedlast(rows_to_delete, info.row - 1)
+    end
+
+    model.name_to_constraint_index = nothing
+
+    # Delete resultant variables
+    resvars = [_info(model, c).resvar for c in cs]
+    MOI.delete(model, resvars)
+
```
odow, line 310:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    _update_if_necessary(model)
+    ret = GRBdelgenconstrs(model, length(rows_to_delete), rows_to_delete)
+    _check_ret(model, ret)
+
+    for (_, info) in model.nl_constraint_info
+        info.row -= searchsortedlast(rows_to_delete, info.row - 1)
+    end
+
+    model.name_to_constraint_index = nothing
+
```
odow, line 306:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    rows_to_delete = sort!([Cint(_info(model, c).row - 1) for c in cs])
+    println(rows_to_delete)
+    _update_if_necessary(model)
+    ret = GRBdelgenconstrs(model, length(rows_to_delete), rows_to_delete)
+    _check_ret(model, ret)
+
+    for (_, info) in model.nl_constraint_info
+        info.row -= searchsortedlast(rows_to_delete, info.row - 1)
+    end
+
```
odow, line 304:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+function MOI.delete(
+    model::Optimizer,
+    cs::Vector{<:MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S}},
+) where {S}
+    rows_to_delete = sort!([Cint(_info(model, c).row - 1) for c in cs])
+    println(rows_to_delete)
+    _update_if_necessary(model)
+    ret = GRBdelgenconstrs(model, length(rows_to_delete), rows_to_delete)
+    _check_ret(model, ret)
+
```
odow, line 300:
```suggestion
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+function MOI.set(
+    model::Optimizer,
+    ::MOI.ConstraintName,
+    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
+    name::String,
+) where {S}
+    _update_if_necessary(model, force = true)
+    info = _info(model, c)
+    info.name = name
+    if !isempty(name)
```
odow, line 337:
```suggestion
    if length(name) <= GRB_MAX_NAMELEN
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+) where {S}
+    return _info(model, c).name
+end
+
+function MOI.set(
+    model::Optimizer,
+    ::MOI.ConstraintName,
+    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
+    name::String,
+) where {S}
```
odow, line 333:
```suggestion
)
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    ::MOI.ConstraintName,
+    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
+) where {S}
+    return _info(model, c).name
+end
+
+function MOI.set(
+    model::Optimizer,
+    ::MOI.ConstraintName,
+    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
```
odow, line 331:
```suggestion
    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction},
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+
+    _require_update(model, model_change = true)
+    return
+end
+
+function MOI.get(
+    model::Optimizer,
+    ::MOI.ConstraintName,
+    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
+) where {S}
```
odow, line 324:
```suggestion
)
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    end
+
+    _require_update(model, model_change = true)
+    return
+end
+
+function MOI.get(
+    model::Optimizer,
+    ::MOI.ConstraintName,
+    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S},
```
odow, line 323:
```suggestion
    c::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction},
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    # Remove resultant variable
+    MOI.delete(model, info.resvar)
+    _require_update(model, model_change = true)
+    return
+end
+
+function MOI.delete(
+    model::Optimizer,
+    cs::Vector{<:MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S}},
+) where {S}
```
odow, line 294:
```suggestion
)
```

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+    model.name_to_constraint_index = nothing
+    # Remove resultant variable
+    MOI.delete(model, info.resvar)
+    _require_update(model, model_change = true)
+    return
+end
+
+function MOI.delete(
+    model::Optimizer,
+    cs::Vector{<:MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S}},
```
odow, line 293:
```suggestion
    cs::Vector{<:MOI.ConstraintIndex{MOI.ScalarNonlinearFunction}},
```

odow (commented):
I probably missed some things, but it looks okay.

It's a nit, but I strongly dislike blank lines in functions. If you need whitespace to make it readable, it says that your function is too large/complicated.

There's also a few Julia-isms, like `isa` and using dispatch. I know the wider codebases in `jump-dev` don't always follow these rules; that's mostly because they have been written and maintained over quite a long period of changing Julia syntax!

There was only one thing that looked obviously wrong to me and needs clarifying, and that's the `if coeff > 1.0` in `_add_expression_tree_variable`. Do you mean `if !isone(coeff)`?

src/MOI_wrapper/MOI_nonlinear.jl
```diff
@@ -0,0 +1,349 @@
...
+# Cases:
+#   1. constant * var
+#   2. +/- var
+#   3. +/- constant * var
+function _check_nonlinear_singlevar(expr)
+    return typeof(expr) == MOI.ScalarNonlinearFunction && (
+        ( # Case 1.
+            expr.head == :* &&
+            length(expr.args) == 2 &&
+            typeof(expr.args[2]) == MOI.VariableIndex
```
odow, line 124: I don't think this actually checks case 1? what if it is variable * variable?

simonbowly: Won't variable * variable arrive in _process_nonlinear as a ScalarQuadraticFunction?

odow: Not always:
```julia
julia> model = Model();

julia> @variable(model, x)
x

julia> @objective(model, Min, @force_nonlinear(x * x))
x * x

julia> model
A JuMP Model
├ solver: none
├ objective_sense: MIN_SENSE
│ └ objective_function_type: NonlinearExpr
├ num_variables: 1
├ num_constraints: 0
└ Names registered in the model
  └ :x

julia> objective_function(model)
x * x

julia> typeof(objective_function(model))
NonlinearExpr (alias for GenericNonlinearExpr{GenericVariableRef{Float64}})

julia> objective_function(model).head
:*

julia> objective_function(model).args
2-element Vector{Any}:
 x
 x
```

simonbowly: Hmm. Given that all the MOI tests are passing (when I run this locally), is it possible some important ones are still being skipped? Or is the coverage just not that good for NLP?

odow: > Or is the coverage just not that good for NLP?

The coverage just isn't as good. Until the introduction of `ScalarNonlinearFunction`, we didn't have a good way of writing `MOI.Test.` for nonlinear. 

There are a bunch of other tests in https://github.com/jump-dev/MINLPTests.jl. See KNITRO for an example of how to call: https://github.com/jump-dev/KNITRO.jl/blob/master/test/MINLPTests/run_minlptests.jl

simonbowly: > https://github.com/jump-dev/MINLPTests.jl

Neat, thanks!

torressa:
Cheers for the review! @odow 

> There was only one thing that looked obviously wrong to me and needs clarifying, and that's the if coeff > 1.0 in _add_expression_tree_variable. Do you mean if !isone(coeff)?

Yes, this was wrong! Thanks!

> It's a nit, but I strongly dislike blank lines in functions. If you need whitespace to make it readable, it says that your function is too large/complicated.

I will have another pass and remove a bunch of these.

src/MOI_wrapper/MOI_nonlinear.jl
torressa: >  Won't variable * variable arrive in _process_nonlinear as a ScalarQuadraticFunction?

I also thought this would be the case. Thanks for double checking this!

torressa: > Hmm. Given that all the MOI tests are passing (when I run this locally)

I think the `test_nonlinear_scalarquadraticfunction` is failing because of this.

simonbowly: Ah, right you are. I added that test thinking it would catch the missing dispatch of ScalarQuadraticFunction in the _process_nonlinear loop. But I see it actually catches this case. That explains why it didn't give the error message I expected!

test/MOI/MOI_wrapper.jl
```diff
@@ -977,6 +988,260 @@ function test_primal_feasible_status()
...
+    # s.t sqrt(*(x, x) + *(y, y)) <= 1
+    # x, y >= 0
+    #
+    # -> x = y = 1/sqrt(2)
+    x = MOI.add_variable(model)
+    y = MOI.add_variable(model)
+    MOI.set(model, MOI.ObjectiveSense(), MOI.MAX_SENSE)
+    f = 1.0 * x + 1.0 * y
+    MOI.set(model, MOI.ObjectiveFunction{typeof(f)}(), f)
+    f1 = MOI.ScalarNonlinearFunction(:*, Any[x, x])
```
odow, line 1235: I made a few changes to the tests. In particular, this used to be `Any[x, y]`, but I think you intended `Any[x, x]`?

odow:
Will be unblocked by https://github.com/JuliaRegistries/General/pull/119224

torressa:
Sorry, CI is failing due to an expired license. I will ask for an extension

odow:
Looks like I managed to trigger CI right on the boundary of when the license key updated, so some jobs failed because of an invalid license, and some jobs failed because of the code issue.

src/MOI_wrapper/MOI_nonlinear.jl
simonbowly: @odow from JuMP's perspective, is there actually any need for this special handling?

i.e. if presented with an explicit nonlinear function `:* {1.0, x}`, is it necessary to discard the parent node? If there's a ScalarAffineFunction or ScalarQuadraticFunction in the tree, it would make sense to come up with a simplified representation where possible, but for explicit nonlinear I'd guess it's preferable to represent it more or less faithfully to the solver.

odow:
I added support for our MINLPTests to https://github.com/jump-dev/Gurobi.jl/pull/586. Merge it into here if/when you want.

I didn't look in detail at the test failures, but it seems like there are still a bunch of bugs in the parser.

src/MOI_wrapper/MOI_nonlinear.jl
odow: I don't know that I understand your question. JuMP doesn't have any perspective. It gives you what the user wrote. And it's up to you to do with that as you please.

I haven't looked in enough detail at why you're doing these calls.

But yes, I generally agree what it's preferable to pass the solver more or less what the user wrote.

simonbowly: For example we can see both of these appear in the processing of a nonlinear function:
1. A ScalarAffineFunction with one term `1.0 * x`
2. A ScalarNonlinearFunction with `head=:*` and two children `Float{1.0}` and `VariableIndex`

I was wondering whether JuMP expects a different representation of those two objects in the underlying solver. Sounds like there are no requirements there.

simonbowly:
> I added support for our MINLPTests to #586. Merge it into here if/when you want.
> 
> I didn't look in detail at the test failures, but it seems like there are still a bunch of bugs in the parser.

Cheers! Yes, let's get this branch passing on its own first.

src/MOI_wrapper/MOI_nonlinear.jl
odow: JuMP doesn't care

simonbowly:
I refactored the expression tree traversal to use type dispatch and sort out a few parent pointer issues. On the API side, everything now looks rosy.

@odow do the changes in [ecb4c74](https://github.com/jump-dev/Gurobi.jl/pull/575/commits/ecb4c74d77343bead6032ab44ec12177f0802580) look reasonable? I would say it makes the stack logic a little easier to follow, though it's a lot of arguments being passed around. Would typical Julia style be to wrap up the common arguments (model, stack, opcode, data, parent) into another type.

Also I guess the function name should be `_add_expression_tree_node!` since it's always modifying its arguments?

odow:
I made a couple of changes. I'll merge this once CI passes. We can address the MINLPTests in a separate PR, and I'll re-check the skipped tests in a new PR as well.

odow (approved):

PR #584, opened 2024-11-01
Fix callback attributes in MOI.ListOfModelAttributesSet

### LINKED ISSUE #579: ListOfModelAttributesSet does not include callbacks
Hey! 

As the title says, we get an error when printing in `LP` format with Gurobi:
```
ERROR: MathOptInterface.UnsupportedAttribute{Gurobi.CallbackFunction}: Attribute Gurobi.CallbackFunction() is not supported by the model.
Stacktrace:
```
Here is a MRE on the latest version `1.3.1` of Gurobi.jl.

```
import Gurobi
import MathOptInterface as MOI

model = MOI.instantiate(() -> Gurobi.Optimizer(); with_cache_type = Float64, with_bridge_type = Float64)
MOI.set(model, Gurobi.CallbackFunction(), () -> nothing)
MOI.Utilities.loadfromstring!(
           model,
           """
       variables: x, y, z
       minobjective: 1.0 * x + 1.0 * y + 1.0 * z
       c1: x + y == 2.0
       c2: x + y + z >= 0.0
       c3: 1.0 * x * x + -1.0 * y * y + -1.0 * z * z >= 0.0
       x >= 0.0
       y >= 0.0
       z >= 0.0
       """)
print_options = (; format = MOI.FileFormats.FORMAT_LP)
dest = MOI.FileFormats.Model(; print_options...)
MOI.copy_to(dest, model)
```

If I omit the line that sets the `Gurobi.CallbackFunction`, printing works. Also using the `Gurobi.Optimizer()` directly instead of a cached optimizer works.

Edit: I thought we could set `CallbackFunction` to `nothing` before printing but that unfortunately resets the optimizer state. Printing works correctly in that case.

Thank you!

odow:
Closes #579

### DIFF
--- a/test/MOI/MOI_callbacks.jl
+++ b/test/MOI/MOI_callbacks.jl
@@ -124,6 +124,9 @@ function test_lazy_constraint_callback()
     @test MOI.get(model, MOI.VariablePrimal(), y) == 2
     attrs = MOI.get(model, MOI.ListOfModelAttributesSet())
     @test MOI.LazyConstraintCallback() in attrs
+    MOI.set(model, MOI.LazyConstraintCallback(), nothing)
+    attrs = MOI.get(model, MOI.ListOfModelAttributesSet())
+    @test !(MOI.LazyConstraintCallback() in attrs)
     return
 end
 
@@ -254,6 +257,9 @@ function test_user_cut_callback()
     @test user_cut_submitted
     attrs = MOI.get(model, MOI.ListOfModelAttributesSet())
     @test MOI.UserCutCallback() in attrs
+    MOI.set(model, MOI.UserCutCallback(), nothing)
+    attrs = MOI.get(model, MOI.ListOfModelAttributesSet())
+    @test !(MOI.UserCutCallback() in attrs)
     return
 end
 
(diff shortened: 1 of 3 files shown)


test/MOI/MOI_callbacks.jl
```diff
@@ -443,6 +470,12 @@ function test_CallbackFunction_callback_LazyConstraint()
     @test Gurobi.GRB_CB_MESSAGE in cb_calls
     @test Gurobi.GRB_CB_PRESOLVE in cb_calls
     @test Gurobi.GRB_CB_MIPSOL in cb_calls
+    attrs = MOI.get(model, MOI.ListOfModelAttributesSet())
+    @test Gurobi.CallbackFunction() in attrs
+    MOI.set(model, MOI.CallbackFunction(), nothing)
```
odow, line 475:
```suggestion
    MOI.set(model, Gurobi.CallbackFunction(), nothing)
```

test/MOI/MOI_callbacks.jl
```diff
@@ -443,6 +470,12 @@ function test_CallbackFunction_callback_LazyConstraint()
...
+    attrs = MOI.get(model, MOI.ListOfModelAttributesSet())
+    @test !(MOI.CallbackFunction() in attrs)
```
odow, line 477:
```suggestion
    @test !(Gurobi.CallbackFunction() in attrs)
```

PR #587, opened 2024-11-13
Update list of excluded nonlinear tests

odow:
Let's see what the damage is

### DIFF
--- a/src/MOI_wrapper/MOI_nonlinear.jl
+++ b/src/MOI_wrapper/MOI_nonlinear.jl
@@ -324,15 +324,30 @@ function MOI.add_constraint(
     opcode = Cint[]
     data = Cdouble[]
     parent = Cint[]
-    sense, rhs = _sense_and_rhs(s)
     _process_nonlinear(model, f, opcode, data, parent)
-    # Add resultant variable
-    vi, ci = MOI.add_constrained_variable(model, s)
-    resvar_index = c_column(model, vi)
+    # Add resultant variable. We don't use MOI.add_constrained_variable because
+    # we don't want it to show up in the bound constraints, etc.
+    #
+    # We do add it to model.variable_info, but these get filtered out from
+    # ListOfVariableIndices etc.
+    vi = CleverDicts.add_item(
+        model.variable_info,
+        _VariableInfo(MOI.VariableIndex(0), 0),
+    )
+    info = _info(model, vi)
+    # Now, set `.index` and `.column`.
+    info.index = vi
+    info.column = _get_next_column(model)
+    lb, ub = _bounds(s)
+    lb = something(lb, -Inf)
+    ub = something(ub, Inf)
+    ret =
+        GRBaddvar(model, 0, C_NULL, C_NULL, 0.0, lb, ub, GRB_CONTINUOUS, "")
+    _check_ret(model, ret)
     ret = GRBaddgenconstrNL(
         model,
         C_NULL,
-        resvar_index,
+        c_column(model, vi),
         length(opcode),
         opcode,
         data,
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -3421,7 +3421,7 @@ function MOI.get(model::Optimizer, ::MOI.NumberOfVariables)
 end
 
 function MOI.get(model::Optimizer, ::MOI.ListOfVariableIndices)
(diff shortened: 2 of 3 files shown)


odow:
These are a blocker for release. I need to look into what's going on. There might be an issue with bridging the scalar nonlinear objective function.

odow:
Ah. I think it might be
https://github.com/jump-dev/Gurobi.jl/blob/729ecb96d8e4b36acac4c503b3eeeed23a53f6ad/src/MOI_wrapper/MOI_nonlinear.jl#L330
which mucks up everything because the user didn't add these variables.

simonbowly:
The first issue I run into if I enable only `test_nonlinear_expression_hs071` is that the test seems to get locked in an infinite loop. But this test should not run at all since:

```julia
MOI.supports(model, MOI.ObjectiveFunction{F}())
```

is false for Gurobi.jl. Further if I just run the body of the test:

```julia
using Test
using Gurobi

MOI = Gurobi.MOI

env = Gurobi.Env()
model = Gurobi.Optimizer(env)

MOI.Utilities.loadfromstring!(
    model,
    """
    variables: w, x, y, z
    minobjective: ScalarNonlinearFunction(w * z * (w + x + y) + y)
    c2: ScalarNonlinearFunction(w * x * y * z) >= 25.0
    c3: ScalarNonlinearFunction(w^2 + x^2 + y^2 + z^2) == 40.0
    w in Interval(1.0, 5.0)
    x in Interval(1.0, 5.0)
    y in Interval(1.0, 5.0)
    z in Interval(1.0, 5.0)
    """,
)
```

then I get the expected error

```
Attribute MathOptInterface.ObjectiveFunction{MathOptInterface.ScalarNonlinearFunction}() is not supported by the model.
```

So is there a testing issue to sort out first? Or is MOI confused by the support for nonlinear constraints but not nonlinear objectives?

torressa:
> So is there a testing issue to sort out first? Or is MOI confused by the support for nonlinear constraints but not nonlinear objectives?

The `@requires` MOI macro works when running in isolation but not when running the test suite 🤷

simonbowly:
> Ah. I think it might be
> 
> https://github.com/jump-dev/Gurobi.jl/blob/729ecb96d8e4b36acac4c503b3eeeed23a53f6ad/src/MOI_wrapper/MOI_nonlinear.jl#L330
> 
> 
> which mucks up everything because the user didn't add these variables.

Your point is for e.g. the hs071 tests, the following does not work as expected due to the added variables?

```
    x = MOI.get(model, MOI.ListOfVariableIndices())
    MOI.set.(model, MOI.VariablePrimalStart(), x, [1.0, 5.0, 5.0, 1.0])
```

So we should be adding the variables directly in the Gurobi API, not through MOI?

odow:
The tests run with a bridge that reformulates the objective into an epigraph constraint:
https://github.com/jump-dev/Gurobi.jl/blob/9720c804e8f52ff316652ca5d1818e7f198a59a6/test/MOI/MOI_wrapper.jl#L49-L50

> So we should be adding the variables directly in the Gurobi API, not through MOI?

Yes, we also need to check that any other bookkeeping related to adding variables/columns is still valid.

odow:
The tests are still running locally, but this is going in the right direction.

simonbowly:
@odow do you need any input from us here? Are the additional variables something that JuMP/MOI can handle, or not?

odow:
The additional variables need to be filtered out of all the various calls so that from MOI's point of view they don't exist.

This also affects variable deletion, etc.

I think the current `master` works for basic build-and-solve examples, but you can probably easily break if you start adding and deleting variables and constraints between solves.

This PR is a good start if you want to keep going. I don't remember where I got up to exactly, but I think it's a blocker for releasing v12, unless we want to mark it as experimental with some known bugs.

simonbowly:
Why add them to the MOI layer if they just need to be filtered out at every step afterwards? Wouldn't it be simpler not to track them at all in MOI and have them only exist in the Gurobi model?

odow:
I think I tried that to begin with, but then it made deletion hard. If you delete the nonlinear constraint you also have to delete the variable and then shift all the indices?

This PR might not be the best approach.

simonbowly:
> I think I tried that to begin with, but then it made deletion hard. If you delete the nonlinear constraint you also have to delete the variable and then shift all the indices?

True, but @torressa sorted out that logic for deletion of regular variables, so I imagine the same approach would work here. I'll give it a go (in a separate PR to avoid breaking your work here).

odow:
Closing in favour of #590

PR #590, opened 2024-11-27
Track nonlinear resultant vars outside of MOI

simonbowly:
Alternative to #587: instead of filtering out the NL resultant variables when querying through MOI, don't add these variables to MOI at all so they don't show up in model queries.

- On deletion of NL constraints, deleted resvar indices are tracked on the deleted list for the next update operation
- In `_update_if_necessary`, the same index tracking trick applied to each entry in model.variable_info is applied to NL constraints as well, so that their resvar indices are appropriately adjusted following deletion of another variable.

### DIFF
--- a/test/MOI/MOI_wrapper.jl
+++ b/test/MOI/MOI_wrapper.jl
@@ -589,6 +589,72 @@ function test_add_constrained_variables()
     return
 end
 
+function test_add_constrained_variable_greaterthan()
+    model = Gurobi.Optimizer(GRB_ENV)
+    MOI.set(model, MOI.Silent(), true)
+    set = MOI.GreaterThan{Float64}(1.2)
+    vi, ci = MOI.add_constrained_variable(model, set)
+    @test MOI.get(model, MOI.NumberOfVariables()) == 1
+    @test MOI.get(model, MOI.ListOfConstraintTypesPresent()) ==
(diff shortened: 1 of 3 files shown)


simonbowly:
This resolves all the tests to do with querying user-visible variables.

test/MOI/MOI_wrapper.jl
```diff
@@ -65,15 +65,14 @@ function test_runtests()
...
-            "vector_nonlinear",
-            "VectorNonlinearFunction",
-            # Tests should be skipped due to RequirementsUnmet, but aren't
-            r"^test_nonlinear_expression_hs071$",
-            r"^test_nonlinear_expression_hs071_epigraph$",
+            "_multiobjective_vector_nonlinear",
+            # Timeouts
             r"^test_nonlinear_expression_hs109$",
             r"^test_nonlinear_expression_hs110$",
+            # MOI.get(MOI.ObjectiveValue()) fails for NL objectives
```
odow, line 72: I'll take a look at this. It's probably on our end.

odow (approved):
I'm happy if you're happy

simonbowly:
I'm happy if @torressa is happy

torressa (approved):
Nice one! Thanks Simon

PR #596, opened 2024-11-29
Add Env(params::Dict) for setting parameters before GRBstartenv

### LINKED ISSUE #588: Provide a way to set configuration parameters in Gurobi.Env
From https://github.com/jump-dev/Gurobi_jll.jl/issues/30.

For WLS licenses, compute servers, cluster managers it's necessary to set parameters on an empty environment before starting it (if, that is, you don't want to keep those parameters in the license file).

https://docs.gurobi.com/projects/optimizer/en/11.0/concepts/environments/configuration.html#empty-environment-example

It would be helpful if there were a way to provide these parameters without calling C API functions directly. In gurobipy we use the following pattern to pass a mapping of parameters to be set before the environment is started:

```python
import gurobipy as gp

env = gp.Env(params={
    "CSManager": "<url>",
    "CSAPIAccessID": "<api key>",
    "CSAPISecret": "<api secret>",
})
```

@odow would it be reasonable to provide a `Gurobi.Env` constructor that takes a [Dict](https://docs.julialang.org/en/v1/base/collections/#Dictionaries)? That would generalise the approach currently in Gurobi.jl of having separate constructors for memory limits and compute servers.

odow:
Closes #588

### DIFF
--- a/test/MOI/MOI_wrapper.jl
+++ b/test/MOI/MOI_wrapper.jl
@@ -1508,6 +1508,20 @@ function test_multiple_solution_nonlinear_objective()
     return
 end
 
+function test_Env()
+    err = ErrorException(
+        "Gurobi Error 10022: Failed to connect to localhost port 61000 after 0 ms: Could not connect to server (code 7, command POST http://localhost:1234/api/v1/cluster/jobs)",
+    )
+    @test_throws err Gurobi.Env("localhost:1234")
+    @test_throws err Gurobi.Env("localhost:1234", "password")
+    @test_throws err Gurobi.Env("localhost:1234", "password"; started = false)
+    env = Gurobi.Env(; output_flag = 2, memory_limit = 1)
+    p = Ref{Cdouble}()
+    @test GRBgetdblparam(env, "MemLimit", p) == 0
+    @test p[] == 1.0
+    return
+end
+
 end  # TestMOIWrapper
 
 TestMOIWrapper.runtests()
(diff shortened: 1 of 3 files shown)


test/MOI/MOI_wrapper.jl
```diff
@@ -1508,6 +1508,25 @@ function test_multiple_solution_nonlinear_objective()
...
+        try
+            f()
+            @assert false
+        catch err
+            @test occursin("Gurobi Error 10022:", err.msg)
+        end
+    end
+    test_err(() ->Gurobi.Env("localhost:1234"))
+    test_err(() -> Gurobi.Env("localhost:1234", "password"))
+    test_err(() -> Gurobi.Env("localhost:1234", "password"; started = false))
```
odow, line 1522:
```suggestion
    test_err(() -> Gurobi.Env("localhost:1234", "password"; started = true))
```

test/MOI/MOI_wrapper.jl
```diff
@@ -1508,6 +1508,25 @@ function test_multiple_solution_nonlinear_objective()
...
+function test_Env()
+    function test_err(f)
+        try
+            f()
+            @assert false
+        catch err
+            @test occursin("Gurobi Error 10022:", err.msg)
+        end
+    end
+    test_err(() ->Gurobi.Env("localhost:1234"))
```
odow, line 1520:
```suggestion
    test_err(() -> Gurobi.Env("localhost:1234"))
```

PR #626, opened 2025-03-14
Change DualObjectiveValue to use Utilities.get_fallback

odow:
@blegat: I'm not really sure what we originally intended here.

```julia
help?> MOI.DualObjectiveValue
  DualObjectiveValue(result_index::Int = 1)

  A model attribute for the value of the objective function of the dual problem for the result_indexth dual result.

  If the solver does not have a dual value for the objective because the result_index is beyond the available solutions (whose number is indicated by the ResultCount attribute), getting this attribute
  must throw a ResultIndexBoundsError. Otherwise, if the result is unavailable for another reason (for instance, only a primal solution is available), the result is undefined. Users should first check
  DualStatus before accessing the DualObjectiveValue attribute.

  See ResultCount for information on how the results are ordered.
```


blegat:
It should be the dual objective evaluated at the dual solution so it's indeed not the objective gap of MIPs

PR #627, opened 2025-03-14
Throw GetAttributeNotAllowed when ObjBound is GRB_INFINITY

odow:
See https://discourse.julialang.org/t/about-jump-is-solved-and-feasible/126868

### DIFF
--- a/test/MOI/MOI_wrapper.jl
+++ b/test/MOI/MOI_wrapper.jl
@@ -1618,7 +1618,10 @@ function test_multiple_solutions_when_dual_infeasible()
     return
 end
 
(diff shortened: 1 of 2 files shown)


test/MOI/MOI_wrapper.jl
```diff
@@ -1618,6 +1618,23 @@ function test_multiple_solutions_when_dual_infeasible()
     return
 end
 
+function test_objbound_errors_when_GRB_INFIINTY()
```
odow, line 1621:
```suggestion
function test_objbound_errors_when_GRB_INFIINTY()
    if Gurobi._GUROBI_VERSION < v"11"
        return
    end
```

PR #635, opened 2025-05-11
Update README.md

WalterMadelim:


odow:
Was there some confusion here? It must be the `::String` `"false"`, not the `::Bool` `false`.

odow:
Closing as won't fix. I don't think we should make this change.

WalterMadelim:
In win11, I set like this.
image

The point here is that you've used the code quote `KEY`, `VALUE` all time. The existing description is `KEY`, `"VALUE"`. There is no need to use a further plain quote like `"VALUE"`, which looks inconsistent.

The description could be 
1. `KEY`, `VALUE`
2. "KEY", "VALUE",
or if you don't find troublesome
3. `"KEY"`, `"VALUE"`

odow:
Did you see the code block below? You don't need to use the Windows GUI.

If we don't use `"`, then there is an ambiguity between `::String` and `::Bool`.

Now admittedly, it doesn't really matter because Julia automatically converts:
```julia
julia> ENV["ABC"] = false
false

julia> ENV["ABC"]
"false"
```


I think the style of unquoted = quoted is fairly standard. See, e.g., https://docs.julialang.org/en/v1/manual/environment-variables/.

It's not even Julia specific. You might do:
```
export GUROBI_JL_USE_GUROBI_JLL="false"; julia -e "import Pkg; Pkg.add(\"Gurobi\")"
```
Unquoted on the left. Quoted on the right.

PR #642, opened 2025-08-06
Mention WLS in installation instructions

odow:
x-ref https://discourse.julialang.org/t/how-do-i-set-up-gurobi-solver-for-jump-with-web-license-service-wls-license/131409

cc @torressa


torressa:
Nice one, thanks Oscar!

PR #649, opened 2025-10-23
Add support for Gurobi 13.0

simonbowly:
@odow we're expecting the final release around Nov 11, at which point we can build Gurobi_jll. This PR uses the headers from the 13.0 beta, which shouldn't see any changes.

I can't recall how this was done for 12.0 - did you merge these header additions into master and release a new Gurobi.jl version only once the JLL was ready?

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -27,7 +27,7 @@ function write_depsfile(path)
 end
 
 const ALIASES =
-    ["gurobi120", "gurobi110", "gurobi100", "gurobi95", "gurobi91", "gurobi90"]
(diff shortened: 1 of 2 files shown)


odow:
Yeah we merged this first (we can make a release), and then made another release with JLL support.

odow (approved):

simonbowly:
Righto. Well, fire when ready, I ran the tests through locally and got no errors. There are no backward compatibility issues with the C API to worry about (that I'm aware of). There are a couple of new features we could take advantage of (some new nonlinear opcodes and an improvement for callbacks) and but I'll open separate issues for those.

PR #668, opened 2026-03-19
WIP: add support for MOI.LazyScalarSet

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/issues/2893


odow:
Closing because we can support this in MathOptLazy.jl instead

PR #670, opened 2026-05-05
Add binary variable bounds

odow:
x-ref https://github.com/jump-dev/MultiObjectiveAlgorithms.jl/pull/198#issuecomment-4380412239

### DIFF
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -1950,14 +1950,14 @@ function MOI.add_constraint(
     ::MOI.ZeroOne,
 )
     info = _info(model, f)
-    ret =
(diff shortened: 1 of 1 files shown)


odow:
This needs some changes to check for existing bounds.

Or we could just decide not to merge this. It's a pretty niche bug in Gurobi that they should fix in an upcoming release.

odow:
I don't think we need this. We can just tell people to do `@variable(model, 0 <= x[1:N, 1:N] <= 1, Bin)`. Adding bounds fixes my original issue in https://github.com/jump-dev/MultiObjectiveAlgorithms.jl/pull/198#issue-4380695066.

If bounds on binary variables are helpful, then Gurobi reeeeeaaaally should be doing this internally.

